PR #151, opened 2017-09-11
Don't pass alwaysBlur down from blurDirty()

dpikt:


dpikt:
@rkillackey can you take a quick look at this? Needed for a project :-)

PR #154, opened 2017-09-19
Pass column name to custom table components

dpikt:


test/tables/sortable-table.test.js
```diff
@@ -90,13 +90,20 @@ test('column can have custom sort function', () => {
 })
 
 test('column can have custom component', () => {
-  const MyCell = () => <td> hi! </td>
+  const MyCell = () => <td> Hi! </td>
```
rkillackey: 💯

dpikt: 😂 I was gonna use this for the test but then I changed it back (or so I thought)

PR #162, opened 2017-10-16
option to not have any maxLength

### LINKED ISSUE #141: Option to ignore maxLength
maxLength for Textarea Input is optional in that there is a default of 500, but there's not a way to turn it off (current workaround: set a very large number)

inveterateliterate:
`maxLength` prop accepts boolean values. Setting this prop to `false` results in unlimited characters, and hides the `LabeledField` with the character count. 

Resolves #141

### DIFF
--- a/src/forms/inputs/textarea.js
+++ b/src/forms/inputs/textarea.js
@@ -8,19 +8,19 @@ import { compose } from '../../utils'
 /**
  *
  * A textarea input that can be used in a `redux-forms`-controlled form. Optionally displays a character count.
- * 
+ *
  * @name Textarea
  * @type Function
  * @param {Object} input - A `redux-forms` [input](http://redux-form.com/6.5.0/docs/api/Field.md/#input-props) object
  * @param {Object} meta - A `redux-forms` [meta](http://redux-form.com/6.5.0/docs/api/Field.md/#meta-props) object
- * @param {Number} [maxLength=500] - The maximum allowed length of the input
+ * @param {Number} [maxLength=500] - The maximum allowed length of the input. Accepts a boolean, e.g., `false` for the option to not set a max length.
  * @param {Boolean} [showCharacterCount=true] - Whether or not to display a character count
  * @example
- * 
+ *
  * function BiographyForm ({ handleSubmit, pristine, invalid, submitting }) {
  *   return (
  *     <form onSubmit={ handleSubmit }>
- *       <Field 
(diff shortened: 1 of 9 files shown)


src/forms/inputs/textarea.js
```diff
@@ -55,17 +55,17 @@ function Textarea (props) {
...
   return (
-    <LabeledField 
+    <LabeledField
       className={ classnames(className, { 'with-character-count': showCharacterCount }) }
       { ...props }
     >
-      { 
-        showCharacterCount &&
+      {
+        maxLength && showCharacterCount &&
```
dpikt: Because of the tricky `&&` thing, a `maxLength` of `0` would render "0" here.

Time to use some weird JS syntax :) https://stackoverflow.com/questions/784929/what-is-the-not-not-operator-in-javascript

src/forms/inputs/textarea.js
```diff
@@ -8,19 +8,19 @@ import { compose } from '../../utils'
...
  *
  * A textarea input that can be used in a `redux-forms`-controlled form. Optionally displays a character count.
- * 
+ *
  * @name Textarea
  * @type Function
  * @param {Object} input - A `redux-forms` [input](http://redux-form.com/6.5.0/docs/api/Field.md/#input-props) object
  * @param {Object} meta - A `redux-forms` [meta](http://redux-form.com/6.5.0/docs/api/Field.md/#meta-props) object
- * @param {Number} [maxLength=500] - The maximum allowed length of the input
+ * @param {Number} [maxLength=500] - The maximum allowed length of the input. Accepts a boolean, e.g., `false` for the option to not set a max length.
```
dpikt: You can put `{Number|Boolean}` as the type. Also, you should just mention `false` rather than saying "a boolean" because the latter suggests that `true` is an acceptable value.

stories/forms/inputs/textarea-story.js
```diff
@@ -33,4 +33,12 @@ storiesOf('Textarea', module)
...
     />
-  ))
\ No newline at end of file
+  ))
+  .add('with no maxLength', () => (
+    <Textarea
+      input={inputProps}
+      meta={{}}
+      maxLength={false}
+      showCharacterCount={false}
```
dpikt: This prop is unecessary

dpikt:
@inveterateliterate merge away when ready!

PR #179, opened 2017-11-10
default maxLength to false

### LINKED ISSUE #175: Make textarea have no max length by default
Seems like we usually don't need it.

inveterateliterate:
Resolves #175

### DIFF
--- a/src/forms/inputs/textarea.js
+++ b/src/forms/inputs/textarea.js
@@ -13,7 +13,7 @@ import { compose } from '../../utils'
  * @type Function
  * @param {Object} input - A `redux-forms` [input](http://redux-form.com/6.5.0/docs/api/Field.md/#input-props) object
  * @param {Object} meta - A `redux-forms` [meta](http://redux-form.com/6.5.0/docs/api/Field.md/#meta-props) object
- * @param {Number} [maxLength=false] - The maximum allowed length of the input. Accepts `false` for the option to not set a max length
+ * @param {Number} [maxLength] - The maximum allowed length of the input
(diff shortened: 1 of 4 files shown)


dpikt:
@inveterateliterate this isn't in the ticket, but since we're making breaking changes- could you change the `showCharacterCount` prop to `hideCharacterCount`?

src/forms/inputs/textarea.js
```diff
@@ -13,8 +13,8 @@ import { compose } from '../../utils'
  * @type Function
  * @param {Object} input - A `redux-forms` [input](http://redux-form.com/6.5.0/docs/api/Field.md/#input-props) object
  * @param {Object} meta - A `redux-forms` [meta](http://redux-form.com/6.5.0/docs/api/Field.md/#meta-props) object
- * @param {Number} [maxLength=500] - The maximum allowed length of the input. Accepts `false` for the option to not set a max length.
- * @param {Boolean} [showCharacterCount=true] - Whether or not to display a character count
+ * @param {Number} [maxLength=false] - The maximum allowed length of the input. Accepts `false` for the option to not set a max length
```
dpikt: Since this prop is omitted by default, it doesn't need a falsey default value- we can just interpret `undefined` the same way as we would `false`. This also means we can change the type to just `Number`, which is less confusing.

dpikt: To be precise- it should still have a defaultValue (to follow the "always specify default if not required" pattern) but it's now an implementation detail that doesn't need to be documented.

PR #196, opened 2018-01-02
Add option groups to `Select` component

### LINKED ISSUE #195: Pass `disabled` key to options in `Select`
I have a use case for Empower Illinois where I need to add a few disabled options, which will act as headers for different sections within the dropdown. Currently we're only passing through the `key` and `value` keys in the `options` array. We could either pass through the `disabled` key or `..rest`.

rkillackey:
Resolves #195 

~Before I update the documentation, I just want to make sure this way of checking whether options are grouped makes sense to you~

### DIFF
--- a/src/forms/helpers/field-prop-types.js
+++ b/src/forms/helpers/field-prop-types.js
@@ -18,23 +18,24 @@ export const fieldOptionsType = PropTypes.arrayOf(
         PropTypes.number
       ]).isRequired
     }),
-    PropTypes.shape({
-      optionGroup: PropTypes.string.isRequired,
-      optionGroupData: PropTypes.arrayOf(
-        PropTypes.oneOfType([
-          PropTypes.string,
-          PropTypes.shape({
-            key: PropTypes.string.isRequired,
-            value: PropTypes.oneOfType([
-              PropTypes.string,
-              PropTypes.number
-            ]).isRequired
-          }),
-        ])
-      )
-    }),
   ])
 )
+
+/**
+ *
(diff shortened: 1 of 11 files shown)


rkillackey:
@dpikt Ready for another look!

src/forms/inputs/select.js
```diff
@@ -85,10 +101,13 @@ function Select (props) {
     className, // eslint-disable-line no-unused-vars
     enablePlaceholderOption,
     options,
+    optionGroups,
     placeholder,
     ...rest
   } = omitLabelProps(props)
-  const optionObjects = serializeOptions(options)
+  const optionObjects = optionGroups.length
```
dpikt: This code is a little hard to read. It might be simpler to define two constants, `optionObjects` and `optionGroupObjects`, and write another helper `serializeOptionGroups()` to create the latter.

src/forms/helpers/field-prop-types.js
```diff
@@ -17,10 +17,26 @@ export const fieldOptionsType = PropTypes.arrayOf(
...
         PropTypes.number
       ]).isRequired
-    })
+    }),
   ])
 )
 
+/**
+ *
+ * A constant representing the `PropTypes` of the `optionGroups` prop for select components, e.g., {@link Select} and {@link CheckboxGroup}
```
dpikt: Shouldn't mention `CheckboxGroup` here since it doesn't have the option group functionality (yet)

rkillackey:
@dpikt Updates made!

PR #203, opened 2018-01-25
Remove `props` from Spinner component

rkillackey:
Resolves `Unknown Prop Warning` on `Spinner` component by not passing down `props`.


dpikt:
@rkillackey instead of passing no props, we could filter non-DOM props using https://www.npmjs.com/package/filter-invalid-dom-props.

rkillackey:
@dpikt Updated!

dpikt:
@rkillackey probably not, but I figured better safe than sorry :)

PR #206, opened 2018-02-08
Allow file-input a limit argument

mrjonesbot:
Why:

On some applications, where file uploads can be large, we need to impose size limits so that bandwidth remains low when fetching uploaded files.


rkillackey:
@mrjonesbot I actually have a similar use case on Artlook, but was able to solve this by creating a custom `FileInput` component, and adding a validation on the file size through `constraints`. LMK if you'd like to pair on that!

dpikt:
@mrjonesbot should this PR be closed?

PR #223, opened 2018-04-03
Default maxLength to null rather than false

inveterateliterate:
Default maxLength to `null` rather than `false` to resolve complaints in console regarding passing 'false' to the dom


dpikt:
Gonna merge this since I'm pushing more changes soon.

PR #225, opened 2018-04-09
Add `DropdownCheckboxGroup` component

### LINKED ISSUE #213: Add DropdownCheckboxGroup component
Relies on https://github.com/LaunchPadLab/lp-hoc/issues/25

rkillackey:
Resolves #213


src/forms/inputs/dropdown-select.js
```diff
@@ -0,0 +1,62 @@
...
+  text: PropTypes.string,
+  selectedValues: PropTypes.arrayOf(PropTypes.string),
+  children: PropTypes.node,
+}
+
+const defaultProps = {
+  selectedValues: [],
+}
+
+// Wraps the `DropdownCheckboxGroup` component
```
dpikt: We might wanna move this to `helpers` since it's not publicly documented and also technically not a redux form input component.

src/forms/inputs/dropdown-select.js
```diff
@@ -0,0 +1,62 @@
...
+}) {
+  return (
+    <div className="dropdown-select">
+      <div className="select-input" onClick={ toggleExpanded }>
+        <p>{ getLabel(selectedValues) }</p>
+      </div>
+      <div 
+        className={ classnames(
+          'options', 
+          'checkboxes',
```
dpikt: This seems specific to the checkbox use case- to make it more extendable we could just include the `className` prop here and pass `'checkboxes'` in from `DropdownCheckboxGroup`

src/forms/inputs/dropdown-select.js
rkillackey: Yeah I was debating that as well. Will do

PR #241, opened 2018-05-29
Allow custom classname on columns

dpikt:
Let me know if you think this is expected behavior, or should be added to the docs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@launchpadlab/lp-components",
-  "version": "3.10.0",
+  "version": "3.10.1",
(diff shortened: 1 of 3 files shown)


dpikt:
@rkillackey sweeet

rkillackey:
@dpikt Any reason we would want to just pass in `...rest`?

dpikt:
@rkillackey I was thinking about that. The problem is that there are some column props that are also DOM props (eg `disabled`) that might lead to some unexpected behavior. If I have to keep writing PRs to add props here I might change that though.

PR #280, opened 2018-09-17
Making flash message container a live region for screen reader support

### LINKED ISSUE #277: Flash message div should have a role of alert
> Sometimes, for example, when using AJAX techniques, the browser is not loading a new page but shows changes, such as form errors, dynamically on the page. The list of errors should be inserted into a prominent container on the top to inform the user in such a case. In addition to the advice above, this container should have the role attribute set to alert to make assistive technology users aware of this change.

https://www.w3.org/WAI/tutorials/forms/notifications/#listing-errors

chawes13:
Resolves #277


dpikt:
@chawes13 feel free to merge this once you bump the version!

PR #302, opened 2018-11-28
Button class name

### LINKED ISSUE #294: Button doesn't update classes correctly if `className` is provided
If you pass `invalid={ true }`, you would expect the button to have class `is-disabled`, but it does not if you've already passed a `className` prop.

chawes13:
Resolves #294


dpikt:
I would consider this to be a patch.

PR #303, opened 2018-11-28
Passing props down to input error

### LINKED ISSUE #301: Input Error doesn't accept class name
You cannot provide an additional class name on an input error. The use case is for additional / different styling for form wide errors that are displayed using a standalone `<InputError />`

https://github.com/LaunchPadLab/lp-components/blob/1b84560a1f24896b11411b9b48e366a3e9cc85c1/src/forms/labels/input-error.js#L55

chawes13:
Resolves #301 

Given that this is new functionality, I think this might qualify as a `minor` update.

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -996,6 +996,8 @@ The error label uses the following rules to determine how it will be displayed:
 -   If the `error` prop is set to a string, the label will display that text
 -   If the `error` prop is set to an array of strings, the label will display those errors separated by commas
 
+In addition to the props below, any extra props will be passed directly to the inner `<span>` element.
+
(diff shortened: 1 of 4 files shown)


dpikt:
Depends whether you think it's expected behavior or not. Try to put yourself back in the headspace of when you created the issue- were you surprised that it didn't work that way or not? 😀

chawes13:
I was surprised but I also didn't read any of the documentation...

In retrospect, patch seems appropriate.

chawes13:
@dpikt Build is passing now that Node version is specified in 3.15.7

PR #304, opened 2018-11-28
Specify Node version

chawes13:
https://yarnpkg.com/lang/en/docs/selective-version-resolutions/

`upath < 1.0.5` is not compatible with Node 10. There is no node version specified in our `package.json`, so Heroku will use the latest LTS (which is now Node 10 as of 10/31/18). I've submitted a PR to `documentation` to bump the `chokidar` dependency, however that will come in `v9` and does not have a confirmed release date.

Alternatively, we could define `node: ^8.0.0` as an engine in our `package.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "bourbon": "^4.3.3",
     "bourbon-neat": "^1.8.0",
     "css-loader": "^0.28.7",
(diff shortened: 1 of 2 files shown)


dpikt:
Looks good. We're probably gonna have to do this a bunch more times :-/

Issue #284, opened 2018-09-25
Dynamically add asterisks to required field when `required` prop is true


inveterateliterate:
related: https://github.com/LaunchPadLab/lp-form/issues/31

inveterateliterate:
can also have a prop that can take in a component for the "required indicator" - reach goal; defaults to asterisk

chawes13:
The asterisk or character that is added should be included in a span that has aria-hidden=true.

```javascript
<label htmlFor="personalDetails.email">Email<span aria-hidden="true">*</span></label>
```

chawes13:
@inveterateliterate Got a target date for this? Would be useful for AZ

inveterateliterate:
is end of year ok @chawes13 ?

chawes13:
Perfect!

inveterateliterate:
@chawes13 do screen readers read that a field is required already  if `required` is true on the input element?

chawes13:
@inveterateliterate yep!

dpikt:
Resolved in #310

PR #310, opened 2019-01-07
add an indicator for required fields

inveterateliterate:
Addresses #284

### DIFF
--- a/src/forms/labels/input-label.js
+++ b/src/forms/labels/input-label.js
@@ -26,7 +26,7 @@ import { toggle, togglePropTypes } from '../../utils'
  * @param {String|Boolean} [label] - Custom text for the label
  * @param {String} [tooltip] - A message to display in a tooltip
  * @param {Boolean} [required] - A boolean value to indicate whether the field is required
- * @param {String|Boolean} [requiredIndicator] - Custom character to denote a field is required (optional, default `*`)
+ * @param {String|Boolean} [requiredIndicator] - Custom character to denote a field is required. Pass false to display no indicator when a field is required. (optional, default `*`)
 
  * @example
  *
@@ -66,7 +66,7 @@ const defaultProps = {
   hint: '',
   label: '',
   tooltip: '',
-  requiredIndicator: '*'
+  requiredIndicator: '*',
 }
 
 function InputLabel ({
@@ -88,7 +88,7 @@ function InputLabel ({
           { labelText }
           {
             required && requiredIndicator &&
-            <span aria-hidden="true">{ requiredIndicator }</span>
+            <span className="required-indicator" aria-hidden="true">{ requiredIndicator }</span>
           }
           {
             hint &&
--- a/src/forms/labels/input-label.js
+++ b/src/forms/labels/input-label.js
@@ -26,7 +26,7 @@ import { toggle, togglePropTypes } from '../../utils'
  * @param {String|Boolean} [label] - Custom text for the label
  * @param {String} [tooltip] - A message to display in a tooltip
  * @param {Boolean} [required] - A boolean value to indicate whether the field is required
- * @param {String|Boolean} [requiredIndicator] - Custom character to denote a field is required. Pass false to display no indicator when a field is required. (optional, default `*`)
+ * @param {String|Boolean} [requiredIndicator] - Custom character to denote a field is required (optional, default `''`)
 
  * @example
  *
@@ -66,7 +66,7 @@ const defaultProps = {
   hint: '',
   label: '',
   tooltip: '',
-  requiredIndicator: '*',
+  requiredIndicator: '',
 }
 
 function InputLabel ({
--- a/src/forms/labels/input-label.js
+++ b/src/forms/labels/input-label.js
@@ -26,7 +26,7 @@ import { toggle, togglePropTypes } from '../../utils'
  * @param {String|Boolean} [label] - Custom text for the label
  * @param {String} [tooltip] - A message to display in a tooltip
  * @param {Boolean} [required] - A boolean value to indicate whether the field is required
- * @param {String|Boolean} [requiredIndicator] - Custom character to denote a field is required (optional, default `''`)
+ * @param {String} [requiredIndicator] - Custom character to denote a field is required (optional, default `''`)
 
  * @example
  *
@@ -58,7 +58,7 @@ const propTypes = {
   name: PropTypes.string.isRequired,
   tooltip: PropTypes.string,
   required: PropTypes.bool,
-  requiredIndicator: PropTypes.oneOfType([ PropTypes.string, PropTypes.bool ]),
+  requiredIndicator: PropTypes.string,
   ...togglePropTypes('tooltipShown')
 }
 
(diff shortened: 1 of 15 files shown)


chawes13:
@inveterateliterate Should the default for `requiredIndicator` be `false`? Otherwise this will impact existing projects when bumping?

src/forms/labels/input-label.js
```diff
@@ -55,34 +57,50 @@ const propTypes = {
...
+  required: PropTypes.bool,
+  requiredIndicator: PropTypes.oneOfType([ PropTypes.string, PropTypes.bool ]),
   ...togglePropTypes('tooltipShown')
 }
 
 const defaultProps = {
   hint: '',
   label: '',
   tooltip: '',
+  requiredIndicator: '*'
```
chawes13: Technically not necessary since undefined is falsey but since the `required` prop is not required, it's best practice to explicitly provide a default here (which would be false)

chawes13: I'm not sure about this but I think this may retroactively add an indicator to existing lp-component inputs that are flagged as required?

src/forms/labels/input-label.js
```diff
@@ -55,34 +57,50 @@ const propTypes = {
...
     <span>
-      {  
+      {
         label !== false &&
         <label htmlFor={ name }>
           { labelText }
-          { 
+          {
+            required && requiredIndicator &&
+            <span aria-hidden="true">{ requiredIndicator }</span>
```
chawes13: Spacing might be an issue here so it will probably help designers to give this a class name (similar to how tooltip has one)

stories/forms/inputs/input.story.js
```diff
@@ -15,33 +15,59 @@ const inputProps = {
...
       meta={{}}
       label={false}
     />
   ))
+  .add('with required true default indicator', () => (
+    <Input
+      input={inputProps}
+      meta={{}}
+      label="Custom Label"
+      required={ true }
```
chawes13: If a prop is true you can just write `required` without explicitly setting it to true. You don't have to change this I think it's more of a personal preference thing

src/forms/labels/input-label.js
inveterateliterate: That's a good point, since we are providing defaults for props not required within the component, but in my mind that prop technically does have a default from the `input` element? The question is can that be lost on someone just looking at this component?

src/forms/labels/input-label.js
inveterateliterate: great idea! do we communicate that these classnames are available somewhere in the docs?

stories/forms/inputs/input.story.js
inveterateliterate: I keep forgetting, thanks for the reminder!

inveterateliterate:
@chawes13 @dpikt it makes sense that this would add the `*` for existing projects if they have `required` as `true` and upgrade this library. That being said, I'm not convinced we should necessarily default the indicator to `false`, as my assumption was that (the larger number) of future projects would find the default `*` useful. That's my view of it, but am ok with either direction if that's what we decide. Can also ask the designers for input if that would be helpful.

src/forms/labels/input-label.js
chawes13: If the `input` element is providing a default, then should we be able to make this prop required?

src/forms/labels/input-label.js
chawes13: Hmm, I can't recall an example of that. Thoughts @dpikt ?

dpikt: We don't usually do that, unless they're dynamic. It's generally fastest to just inspect the element when styling.

chawes13:
@inveterateliterate @dpikt - My personal design philosophy is to indicate when a field is optional as opposed to required and to minimize or eliminate the fields that are optional.

A majority of the projects I've worked on so far have followed this philosophy as opposed to marking fields as required. I don't have the context for the rest of our projects though so we should probably ask @marchwiany!

marchwiany:
@chawes13 I think I've been leaning toward using optional lately on projects because I was noticing entire forms that would need a required asterisk after the labels or if there was a field that wasn't required it would only be one or two fields per form. It helps keep the forms a little cleaner and solves the problem of how not every user knows what the "*" means.

If there was a case where the forms we were building was more split with half required and half optional I might lean more toward using the asterisk or maybe even consider designing the form differently.

chawes13:
@marchwiany If I'm interpreting your comment correctly, you are suggesting that we **should not append** the asterisks (*) by default. That aligns with my interpretation of our design philosophy / forms in general.

@inveterateliterate would you mind making this change? Otherwise I think everything else is good to go!

src/forms/labels/input-label.js
```diff
@@ -55,34 +57,50 @@ const propTypes = {
   label: PropTypes.oneOfType([ PropTypes.string, PropTypes.bool ]),
   name: PropTypes.string.isRequired,
   tooltip: PropTypes.string,
+  required: PropTypes.bool,
+  requiredIndicator: PropTypes.oneOfType([ PropTypes.string, PropTypes.bool ]),
```
dpikt: You can probably remove the boolean type since it's now hidden by default.

stories/forms/inputs/input.story.js
```diff
@@ -15,33 +15,59 @@ const inputProps = {
...
+      requiredIndicator={ '@@@' }
+    />
+  ))
+  .add('with required true no indicator', () => (
+    <Input
+      input={inputProps}
+      meta={{}}
+      label="Custom Label"
+      required
+      requiredIndicator={ false }
```
dpikt: Same comment as above

stories/forms/inputs/input.story.js
```diff
@@ -15,33 +15,59 @@ const inputProps = {
...
+      required
+    />
+  ))
+  .add('with required true custom indicator', () => (
+    <Input
+      input={inputProps}
+      meta={{}}
+      label="Custom Label"
+      required
+      requiredIndicator={ '@@@' }
```
dpikt: Maybe make this an * now?

stories/forms/inputs/input.story.js
```diff
@@ -15,33 +15,59 @@ const inputProps = {
...
   ))
   .add('with no label', () => (
-    <Input 
+    <Input
       input={inputProps}
       meta={{}}
       label={false}
     />
   ))
+  .add('with required true default indicator', () => (
```
dpikt: This is probably more confusing than helpful now that it defaults to hidden.

test/forms/labels/input-label.test.js
```diff
@@ -38,3 +38,13 @@ test('when tooltip provided - toggle tooltip', () => {
   wrapper.find('span.tooltip-trigger').simulate('click')
   expect(wrapper.find('div.tooltip-content.is-active').exists()).toEqual(false)
 })
+
+test('when required true - no custom indicator provided', () => {
```
dpikt: Can you put the expected result in the name of the test? Eg. `when no custom required indicator provided, do not show required indicator`

PR #316, opened 2019-01-27
Include value onBlur

### LINKED ISSUE #315: DateInput clears value when normalize is defined
Redux-Form expects value onBlur when using normalize
https://github.com/erikras/redux-form/issues/2768#issuecomment-292770517
https://github.com/erikras/redux-form/issues/3330#issuecomment-354694471

chawes13:
This resolves #315 

**Disclaimer**: I wasn't sure how to test this. I've tested the fix locally on a client project and it resolves the issue that I'm experiencing. Our jest tests currently treat `redux-form` as an implementation detail and don't test it directly...but that means that simulating `change` events directly on the input doesn't trigger `redux-form` lifecycle events like `normalize` and `format`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@launchpadlab/lp-components",
-  "version": "3.16.0",
+  "version": "3.16.1",
   "engines": {
     "node": "^8.0.0"
   },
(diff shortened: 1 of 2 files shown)


chawes13:
@dpikt I'm not seeing this issue on another DateInput that I'm using in this repo for selecting ticket date. It seems to be a pretty special one-off case with the combination of `parse`, `normalize`, and `format` that I'm using. I created a component locally with this change to see if it impacted the existing input and observed no changes in its behavior.

dpikt:
Good to merge at will!

PR #326, opened 2019-03-27
Fix radio group behavior

### LINKED ISSUE #292: RadioGroup does not meet ARIA authoring practices
https://www.w3.org/TR/wai-aria-practices-1.1/#radiobutton

Main issues:
- If an item is selected, it should be the first element that gets focused
- Arrow keys should control navigation between choices

chawes13:
When using a keyboard to navigate into a radio group, tab will now default to the selected value (or first value if none are selected). Arrow keys can then be used to navigate between choices. Keyboard interaction behaves as expected as defined in the [ARIA Authoring Practices](https://www.w3.org/TR/wai-aria-practices-1.1/#radiobutton).

I put in a valiant effort to try and test the focus issue but that seems virtually impossible with enzyme. Let me know if you've had success elsewhere.

This resolves #292

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -1050,6 +1050,7 @@ Note: When using third party form libraries (e.g., [Redux Form][150]), it's like
 ### Parameters
 
 -   `name` **[String][131]** The name of the associated input
+-   `id` **[String][131]** The id of the associated input (defaults to name) (optional, default `name`)
 -   `hint` **[String][131]?** A usage hint for the associated input
 -   `label` **([String][131] \| [Boolean][133])?** Custom text for the label
 -   `tooltip` **[String][131]?** A message to display in a tooltip
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@launchpadlab/lp-components",
-  "version": "3.16.3",
+  "version": "3.17.0",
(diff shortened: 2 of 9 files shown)


dpikt:
I think this may cause a regression in the label behavior- right now it picks its `for` target based on the name rather than the id of the associated input.

dpikt:
You can test in storybook by clicking the radio input labels.

chawes13:
Shoot. I have another branch where I'm potentially going to rip out the `<Input />` for each individual radio button as it's a bit overkill to have those wrapped in a `<LabeledField />`. The error messages should be at the group level. This would likely be a "breaking" change as style would be impacted across apps.

In the interim, I think I could pass in `id` as a prop on the `input` and modify `<Input />` to take `id || name`?

chawes13:
@dpikt I've proposed a code fix. Good catch

chawes13:
@dpikt Tests added!

chawes13:
Using `name` as the `id` for checkboxes should be fine (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox). I don't think there are any changes required with our implementation.

dpikt:
@chawes13 LGTM!

Issue #271, opened 2018-07-30
Add CI to keep track of bundle size


dpikt:
https://github.com/siddharthkp/bundlesize

PR #332, opened 2019-04-03
Remove moment dependency and add CI for bundle size

### LINKED ISSUE #273: Try a new datepicker library
[React-datepicker](https://github.com/Hacker0x01/react-datepicker) seems abandoned by its creators. And it has a gross `moment` dependency.

https://www.npmjs.com/package/react-date-picker might be a good alternative. We just need to make sure it works with our default styling.

dpikt:
Resolves #273 and #271 

Removes `moment` dependency by upgrading to `react-datepicker` v2. It looks like that dependency switch was the only breaking change between versions 1 and 2 but we should still make sure to QA thoroughly to make sure this remains a minor change for our lib.


dpikt:
Hmm. It looks good in storybook. Unit testing this component has historically been pretty annoying. Potentially a way forward would be to `yarn link` it on a date-heavy app (SC?) a smoke test it there?

dpikt:
Closing for now 😢

PR #339, opened 2019-04-19
Disable click events

### LINKED ISSUE #338: Do not allow keyboard or click events for submitting / disabled buttons
While CSS styling often has `pointer-events: none` set for submitting, if using a keyboard it's still possible to press the button. `Redux-Form` might protect against this but we should override this with a `noop` by default

chawes13:
Resolves #338

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -313,7 +313,8 @@ Type: PropTypes
 
 A simple button component that can be used independently, or as part of a form.
 
-Conditionally adds classes and/or becomes disabled depending on passed props.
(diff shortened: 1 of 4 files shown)


dpikt:
This might be considered a patch 🤔

chawes13:
Yeah... I think technically it could be considered a bug. Want me to switch the base?

dpikt:
👍

chawes13:
Rebased and ready! Can you give me a quick sanity check that things aren't wildly wrong?

PR #344, opened 2019-05-02
Pass props to label/error components and filter invalid properties on native inputs

chawes13:
… documentation


chawes13:
The remaining props don't get passed from labeled field to the input label. Introducing this would require some more investigation re: what is actually passed down from a redux-form field. In the interim, I've removed the inappropriate example from the documentation and cleaned up the `LabeledField` story to demonstrate how to add actions to a custom label component.

dpikt:
@chawes13 Hmm, I'd expect all the props to get passed down to the label component (aka this is a bug). Adding that functionality should be fine as long as we filter invalid props on the default component right?

chawes13:
@dpikt Picking this up from a lonnnnnng time ago (use case just came up on Artlook). I agree, we should pass all props down. I was thinking there might be some edge cases around `aria-` attributes but thinking through some use cases didn't seem to raise any red flags

chawes13:
but I haven't implemented that yet 😆

dpikt:
Haha I didn't catch that. Consider it a major vote of confidence!

chawes13:
Despite the vote of confidence, would still appreciate if you could take another look!

dpikt:
This just fixed a bug for me on AZ!

dpikt:
(thought you might like to know)

chawes13:
🔥 donate to my patreon

dpikt:
`the author of lp-components is looking for a good job -)`

PR #349, opened 2019-06-11
Pass className down to label

### LINKED ISSUE #340: InputLabel doesn't accept className
Now that label components can be passed in through `labelComponent` to an input, I've found that I'm using `InputLabel` on its own more often. I want to directly apply a `className`, but it isn't getting passed down.

dpikt:
Resolves #340

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@launchpadlab/lp-components",
-  "version": "3.20.3",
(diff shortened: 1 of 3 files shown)


dpikt:
I was thinking about it, then I got paranoid and didn't follow through since there's more potential for breaking changes. If we did that I might include it in a major version, since who knows what we're passing down in our code.

PR #379, opened 2019-10-02
Add upload handlers to cloudinary input

### LINKED ISSUE #378: Add props for directly accessing upload response
Since `CloudinaryInput` is running the upload `onChange`, `redux-form` doesn't catch the exception that is thrown. It results in an unhandled rejected promise. How can we access / handle this error?

chawes13:
Resolves #378


chawes13:
I wrote the tests!!!

PR #380, opened 2019-10-08
Modify File Input to accept multiple files

### LINKED ISSUE #346: FileInput doesn't support multiple files
In the past, we've used a `FieldArray` to cover this case, but that's not ideal. Theoretically, you could add a `multiple` prop, (https://www.w3schools.com/tags/att_input_multiple.asp) but I'm not sure how that would play with redux-form.

chawes13:
Request for comments on initial approach before diving further into refactoring `CloudinaryFileInput` and potentially adding `react-dropzone`

**Punchlist**:
- [x] Add tests
- [x] Add stories
- [x] Update documentation
- [x] Refactor using hooks
- [x] Fix React `act()` warning in Cloudinary File Input tests (https://share.getcloudapp.com/nOuvxBve)
- [x] Get 100% code coverage (`yarn jest test/forms/inputs/file-input.test.js --collectCoverageFrom="src/forms/inputs/file-input/file-input.js" --coverage`)
- [x] Update migration guide
- [x] Test with `redux-form`
- [x] Test with existing initial values
- [x] Test with `Formik`

Resolves #346


chawes13:
@dpikt Included a POC for forcing `onChange` every time but filtering based on values held internally. Since `FileList` is read-only, we are essentially maintaining our own list via `state`.

Additionally, I added in a comment about `onLoad`. At the very least, I think setting state needs to be dependent on the successful fulfillment of that promise. Otherwise, we aren't keeping the internal "File List" up-to-date. I haven't quite figured out how to make `onChange` work in support of `multiple` vs `single`

src/routes/index.js
```diff
@@ -1,2 +1,2 @@
-export AuthorizedRoute from './authorized-route'
-export UnauthorizedRoute from './unauthorized-route'
+export { default as AuthorizedRoute } from './authorized-route'
+export { default as UnauthorizedRoute } from './unauthorized-route'
```
chawes13, line 2: This fixed a linter warning

src/forms/inputs/file-input/file-input.js
```diff
@@ -48,107 +59,203 @@ import classnames from 'classnames'
...
+      </div>
+    </LabeledField>
+  )
+}
+
+// Do not reload files that have been successfully loaded
+function removeExistingFiles (newFiles, existingFiles) {
+  return newFiles.filter((file) => {
+    return !existingFiles.some(({ name, lastModified }) => {
+      return name === file.name && lastModified === file.lastModified
```
chawes13, line 216: I think this is a decent proxy for a file being the "same"

src/forms/inputs/file-input/file-input.js
```diff
@@ -1,34 +1,45 @@
...
  * @param {Object} meta - A `redux-forms` [meta](http://redux-form.com/6.5.0/docs/api/Field.md/#meta-props) object
- * @param {Function} [onLoad] - A callback fired when the file is loaded
+ * @param {Function} [readFiles] - A callback that is fired with new files and is expected to return an array of file objects with the `url` key set to the "read" value. This can be either a data URL or the public URL from a 3rd party API
+ * @param {Boolean} [multiple=false] - A flag indicating whether or not to accept multiple files
+ * @param {Function} [onRemove=noop] - A callback fired when the file is removed (only available when `multiple` is set to `true`)
+ * @param {Function} [removeComponent=RemoveButton] - A custom component that receives the `onRemove` callback (only available when `multiple` is set to `true`)
  * @param {String} [thumbnail] - A placeholder image to display before the file is loaded
- * @param {Boolean} [hidePreview] - A flag indicating whether or not to hide the file preview
+ * @param {Boolean} [hidePreview=false] - A flag indicating whether or not to hide the file preview
```
danparnella, line 40: Should `previewComponent` be listed here as a prop option?

stories/forms/inputs/file-input.story.js
```diff
@@ -40,4 +40,11 @@ storiesOf('FileInput', module)
       meta={{}}
       previewComponent={FilenamePreview}
```
danparnella, line 41: `FilenamePreview` needs to be updated to take in `value` instead of `file`.

danparnella (changes requested):
Just a few small things, but loving the multi-file story, working great and definitely something we can use on artlook!

src/forms/inputs/file-input/file-input.js
danparnella: I also just had a chat with Mike who wasn't aware that you could pass `accept` with a list of the only accepted filetypes. I know that's a default `input` prop, but could be helpful to make everyone aware it's available with this input.

migration-guides/v6.0.0.md
```diff
@@ -1,24 +1,28 @@
...
 This version contains the following breaking changes:
 
 1. `<Select />` now has a `placeholder` prop by default ("Select")
 1. `<Button />` and `<SubmitButton />` now use `aria-disabled` instead of `disabled`
 1. The `style` prop on `<Button />` and `<SubmitButton />` has been renamed to `variant`
 1. Labels for `<CheckboxGroup />` and `<RadioGroup />` now use a `<legend>` html tag rather than a `<label>`
 1. The `hideLabel` prop on `<RangeInput />` has been renamed to `hideRangeValue`
+1. `<CloudinaryFileInput />` and `<FileInput />` components now store an object (or array of objects) representing the file data as opposed to a string
+1. `<CloudinaryFileInput />` and `<FileInput />` components no longer accept an `onLoad` prop
+1. The `previewComponent` for a file input no longer receives a `file` prop and `value` is a file object (or array of file objects) as opposed to a string
```
bhennes2, line 11: Does it make it easier to have the same interface for the `value` to always be an array and just an array of one object in the non-multi file support scenario?

bhennes2 (commented):
Looking good to me but did have one comment around whether or not the value should always be an `array` type.

Admittedly I'm not intimately familiar with our `FileInput` component currently so only reliably able to offer more high-level comments.

migration-guides/v6.0.0.md
chawes13: That would match the native [FileList](https://developer.mozilla.org/en-US/docs/Web/API/FileList) implementation

danparnella: Sorry! The edit didn't come through email, but yeah, I like the simplification (just always being an array). Taking a look at these updates now.

src/forms/inputs/file-input/file-input.js
```diff
@@ -11,22 +11,20 @@ import {
...
 import ImagePreview from './image-preview'
-import { first, noop, generateInputErrorId, isString, removeAt } from '../../../utils'
+import { noop, generateInputErrorId, isString, removeAt } from '../../../utils'
 import classnames from 'classnames'
 
 /**
- * A file input that can be used in a `redux-forms`-controlled form.
- * The value of this input is a file object or an array of file objects with the `url` set to the base64 encoded data URL of the loaded file(s).
+ * A file input that can be used in a `redux-form`-controlled form.
+ * The value of this input is an array of file objects, with the `url` set to the base64 encoded Ddata URL of the loaded file(s) by default.
```
danparnella, line 19: Extra `D` in `data`

docs.md
```diff
@@ -569,27 +573,32 @@ export default TodoForm
 
 ## FileInput
 
-A file input that can be used in a `redux-forms`-controlled form.
-The value of this input is the data URL of the loaded file.
+A file input that can be used in a `redux-form`-controlled form.
```
danparnella, line 576: Should we make this change (`redux-forms` to `redux-form`) globally?

src/forms/inputs/file-input/file-input.js
```diff
@@ -48,107 +60,195 @@ import classnames from 'classnames'
...
+    const { onChange, value } = input
+    const [removedFile, remainingFiles] = removeAt(value, idx)
+
+    try {
+      await onRemove(removedFile)
+
+      // If all files have been removed, then reset the native input
+      if (!remainingFiles.length) clearFileInput()
+
+      // This method is only available when multiple="true", so always set an array
```
danparnella, line 132: Is this comment still applicable (since we're now **always** setting the value to an array)?

src/forms/inputs/file-input/file-input.js
```diff
@@ -1,34 +1,46 @@
...
- * This callback will be passed the data URL of the file, as well as the `File` object itself.
- *
- * By default, this component displays a thumbnail preview of the loaded file. This preview can be customized
+ * By default, this component displays a thumbnail preview of the loaded file(s). This preview can be customized
  * by using the `thumbnail` or `hidePreview` props, as well as by passing a custom preview via `previewComponent` or `children`.
  *
  * A component passed using `previewComponent` will receive the following props:
- * - `file`: the uploaded file object, or `null` if no file has been uploaded.
- * - `value`: the current value of the input (data URL or empty string)
+ * - `value`: the current value of the input (an array of file objects)
```
danparnella, line 27: I think we swapped this back to `file` as the prop for the `previewComponent`.

src/forms/inputs/file-input/file-input.js
```diff
@@ -48,107 +60,195 @@ import classnames from 'classnames'
...
 }
 
-FileInput.propTypes = propTypes
+function RemoveButton ({ file, onRemove }) {
+  return (
+    <button
+      type="button"
+      className="remove-file"
+      onClick={onRemove}
+      aria-label={`Remove ${file.name}`}
```
danparnella, line 244: Need optional chaining here

PR #383, opened 2019-10-09
Make paginator accessible

chawes13:
Resolves: https://app.asana.com/0/inbox/1109299621236523/1143554707945654/1143988179303738


src/controls/paginator/page-link.js
```diff
@@ -7,19 +7,36 @@ const propTypes = {
...
 const defaultProps = {
   className: '',
   active: false,
-  onClick: noop
+  onClick: noop,
+  'aria-label': '',
 }
 
-function PageLink ({ className, active, onClick, children }) {
+const triggerOnEnter = (fn) => {
```
dpikt: Util? 😄

src/controls/paginator/page-link.js
```diff
@@ -7,19 +7,36 @@ const propTypes = {
...
+function PageLink ({ className, active, onClick, children, 'aria-label': ariaLabel }) {
   return (
-    <li className={classnames(className, { 'active': active })}>
-      <a onClick={ onClick }>
+    <li className={ classnames(className, { 'active': active }) }>
+      <a
+        onClick={ onClick }
+        onKeyPress={ triggerOnEnter(onClick) } // keyboard interaction requirement
+        aria-current={ active ? 'page' : false }
+        aria-label={ ariaLabel }
```
dpikt: Use `...rest` for this?

src/controls/paginator/paginator.js
```diff
@@ -58,24 +57,28 @@ const defaultProps = {
   nextLabel: 'Next'
 }
 
-function Paginator ({ 
+const Nav = ({ children }) => <nav className="pagination" aria-label="pagination">{children}</nav>
```
dpikt: `function` keyword is better for devtools (gives it a displayName)

chawes13:
@dpikt Changes are in!

src/utils/trigger-on-keys.js
```diff
@@ -0,0 +1,24 @@
...
+ * @returns {Function} - Returns a function that takes an event and watches for keys
+ * 
+ * @example
+ * 
+ * const triggerOnEnter = triggerOnKeys(() => console.log('Hi'), [13])
+ * function MyExample () { return <Example onKeyPress={triggerOnEnter} /> }
+ */
+
+function triggerOnKeys (fn, keyCodes) {
+  const codes = Array.isArray(keyCodes) ? keyCodes : [keyCodes]
```
dpikt: Can use `castArray` from `lodash` 😀

src/utils/trigger-on-keys.js
```diff
@@ -0,0 +1,24 @@
...
+  return function (e) {
+    const key = e.which || e.keyCode
+    if (!codes.some((keyCode) => keyCode == key)) return
+    
+    return fn()
```
dpikt: Might make sense to pass the event through here.

src/controls/paginator/page-link.js
```diff
@@ -1,25 +1,31 @@
...
 
-function PageLink ({ className, active, onClick, children }) {
+function PageLink ({ className, active, onClick, children, ...rest }) {
   return (
-    <li className={classnames(className, { 'active': active })}>
-      <a onClick={ onClick }>
+    <li className={ classnames(className, { 'active': active }) }>
+      <a
+        onClick={ onClick }
+        onKeyPress={ triggerOnKeys(onClick, 13) } // keyboard interaction requirement
```
dpikt: Using a constant for the key code might make this line more readable

chawes13:
@dpikt would you consider the switch from `div` to `nav` a breaking change?

chawes13:
@dpikt it's strange because this feels like it could be breaking (div to nav) or it could be patch (all impl. details and aria), but probably is better to be a minor with a call out to check styles?

@marchwiany Should we also review this in the default style guide in the client template to make sure those styles are up-to-date?

chawes13:
@dpikt should I export the util?
```
export {
  compareAtPath,
  serializeOptions,
  serializeOptionGroups,
  stripNamespace,
} from './utils'
```

marchwiany:
@chawes13 yeah I can definitely update styles in the client template!

dpikt:
@chawes13 I'm totally impartial on exporting the util. If all the other utils are being exported, might be nice to keep it consistent, but I don't imagine it getting used much at all.

As far as the version goes, if the client template styles don't break with the change, we can keep it a patch. If they do, it will be breaking.

Speaking of- at some point I'd like to do more thinking about how these components interact with styling in general. Our current approach seems kinda brittle/redundant.

chawes13:
@marchwiany Does the latest client template have a paginator component in the default styleguide or would we need to add that?

chawes13:
Paginator in client template pre-update:
image

Paginator in client template post-update:
image

Marking this as a minor update

PR #384, opened 2019-10-09
a11y group legends

chawes13:
Adding this in v4 as this change will likely cause a visual regression in any app we have that hasn't explicitly assigned `legend`s to have visual parity with labels.


inveterateliterate:
and if so what's the release plan for v4? we can ignore this in our CI until then

PR #311, opened 2019-01-09
legacy-v4 [on hold]

dpikt:
Resolved issues:
* #293 
* #233 
* #325 
* Add legends for groups (PR #384)

PR #386, opened 2019-10-10
Make TabBar Accessible

### LINKED ISSUE #345: Can't tab into TabBar or Paginator
Keyboard interactions don't work as expected with these components.

chawes13:
https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel

Resolves #345


dpikt:
@chawes13 I don't think the `defaultValue` prop is necessary (the default to first tab makes sense), if another default is desired it can accomplished by using `value`.

It might be an interesting enhancement to make this input "controllable": https://medium.com/myheritage-engineering/how-controllable-react-components-maximize-reusability-86e3d233fa8e. If we want to keep it controlled, though, we should probably make `value` required.

dpikt:
That being said, I'm pretty sure that React docs article I linked in #388  was about NOT doing precisely what this article suggests... 🤔

src/controls/tab-bar/focus.js
```diff
@@ -0,0 +1,81 @@
+const PERTINENT_KEY_CODES = {
+  DOWN: '40',
+  UP: '38',
+  HOME: '36',
+  END: '35',
+  LEFT: '37',
+  RIGHT: '39',
+}
+
+function getClosestTabList (el) {
```
dpikt: Can you add some brief comments about what the functions in this file do?

chawes13:
> That being said, I'm pretty sure that React docs article I linked in #388 was about NOT doing precisely what this article suggests... 🤔

@dpikt Can we meet on this? It's a little over my head at the minute

chawes13:
@dpikt this is ready for review! I bumped it as a `minor` version given the newly added behavior to default to the first tab. The other fixes would probably fit under `patch` - it's a grey area.

One thing that seemed really weird to me is that `options` is an optional argument. Should we modify this in #311 ?

src/controls/tab-bar/focus.js
```diff
@@ -0,0 +1,90 @@
...
+function getSafeIndex (length, newIndex) {
+  const maxIndex = length - 1
+  
+  if (newIndex < 0) return getSafeIndex(length, length + newIndex)
+  if (newIndex > maxIndex) return getSafeIndex(length, length - newIndex)
+  
+  return newIndex
+}
+
+function createFocusListener (alignment='horizontal') {
```
dpikt: @chawes13 can you comment this one too? 😇

chawes13: Doneso!

chawes13:
@dpikt LMK if you have any more feedback

src/controls/tab-bar/focus.js
```diff
@@ -0,0 +1,93 @@
...
+function focusExtremeControl (control, top=true) {
+  const tabList = getClosestTabList(control)
+  const tabs = Array.from(tabList.querySelectorAll('[role="tab"]'))
+  
+  top ? tabs[0].focus() : tabs[tabs.length - 1].focus()
+}
+
+// Recursively calculates a "safe" index by wrapping around when the array bounds are exceeded
+// e.g., getSafeIndex([1, 2, 3].length, 3) -> 0
+function getSafeIndex (length, newIndex) {
```
dpikt: Can use the modulus operator for this! `safeIndex = index % array.length`

src/controls/tab-bar/focus.js
```diff
@@ -0,0 +1,93 @@
...
+  UP: '38',
+  HOME: '36',
+  END: '35',
+  LEFT: '37',
+  RIGHT: '39',
+}
+
+// Recursively searches for the closest parent tab list
+function getClosestTabList (el) {
+  return el && (el.matches('[role="tablist"]') ? el : getClosestTabList(el.parentElement))
```
dpikt: This double conditional is a little hard to read.

src/controls/tab-bar/focus.js
```diff
@@ -0,0 +1,93 @@
...
+    
+    const key = (e.which || e.keyCode || '').toString()
+    switch (key) {
+      case PERTINENT_KEY_CODES.DOWN:
+      case PERTINENT_KEY_CODES.UP: {
+        if (!vertical) break
+        
+        e.preventDefault()
+        const direction = (key === PERTINENT_KEY_CODES.DOWN) ? 1 : -1
+        focusNextControl(e.target, direction)
```
dpikt: These arguments are very flag-like, I'd suggest creating separate functions and DRYing within the functions themselves:
```js
focusNextControl()
focusPreviousControl()
focusBottomControl()
focusTopControl()
```

src/controls/tab-bar/focus.js
chawes13: Does that handle the case where the index is negative? e.g., `fn([0, 1, 2], -1) -> 2`

dpikt: Looks like you need a floored modulo - so a helper is still necessary ☹️ https://stackoverflow.com/a/43827557

chawes13: 🤯

dpikt: By the way, ruby (and lots of other languages) would do that by default. Also I found this neat function too (it doesn't wrap though): https://lodash.com/docs/4.17.15#nth

src/controls/tab-bar/focus.js
chawes13: Thanks for the feedback! I've incorporated this into `focus.js` and it has significantly improved the quality of the code imo. If you have time, please take a look and let me know if you would have done anything differently!

PR #388, opened 2019-10-11
Refactor to a controlled file input

chawes13:
- Overall feedback on approach
- Error handling strategy?
- Syntax, etc.


chawes13:
@dpikt Let me know if anything jumps out at you as a must fix/change before I port this over to a standalone project

src/forms/inputs/cloudinary-file-input.js
```diff
@@ -53,25 +52,40 @@ const defaultProps = {
...
+        try {
+          const uploadFilePromises = files.map(async (file) => {
+            const cloudinaryRes = await upload(file.url, file)
+            return mapCloudinaryResponse(file, cloudinaryRes)
+          })
+          
+          const uploadedFiles = await Promise.all(uploadFilePromises)
+          onUploadSuccess(uploadedFiles)
+          return uploadedFiles
+        } catch (e) {
```
dpikt: You'll need to rearrange this `catch` to match the previous behavior (don't catch exceptions from `onUploadSuccess`)

src/forms/inputs/file-input/helpers.js
```diff
@@ -0,0 +1,49 @@
+import { castArray } from '../../../utils'
+
+export async function readFiles ({ newFiles, existingFiles }) {
+  // Do not reload files that have been successfully loaded
+  const filesToLoad = newFiles.filter((file) => {
```
dpikt: I'd lift this comparison out of this function back into the component (could use a helper for it though)

```
const files = [ ...e.target.files ] // spread might not be necessary
const newFiles = removeExistingFiles(files, values)
const filesWithUrls = await readFiles(newFiles)
```

src/forms/inputs/file-input/file-input.js
```diff
@@ -49,141 +50,133 @@ import { noop, generateInputErrorId, removeAt } from '../../../utils'
...
-      onChange([...value, fileData])
-      this.setState((state) => {
-        return { files: [ ...state.files, file ] }
-      })
-    } else {
-      onChange(fileData)
-      this.setState({ files: [file] })
+    try {
+      // Only add value to form if file(s) successfully load
+      const filesToAdd = await Promise.resolve(onRead(fileValues)) // wrap in a promise (just in case)
```
dpikt: Since `onRead` now corresponds directly with `onChange`, we can probably accomplish the same effect by just using `onChange`. So CloudinaryInput would look like:
```jsx
...
return (
 <FileInput
      input={{ 
        ...input, 
        onChange: async (files) => {
           // upload, then call input.onChange with the result.
        }
}
```
This should reduce the API surface area and make the return value clearer.

src/forms/inputs/file-input/file-input.js
```diff
@@ -49,141 +50,133 @@ import { noop, generateInputErrorId, removeAt } from '../../../utils'
...
-    const [removedFile, remainingFiles] = removeAt(this.state.files, idx)
+  async removeFile (idx) {
+    const { input: { onChange, value }, multiple, onRemove } = this.props
+    const [removedFile, remainingFiles] = removeAt(value, idx)
     
-    onChange(remainingValues)
-    onRemove(removedValue, removedFile)
-    this.setState({ files: remainingFiles })
+    try {
+      await Promise.resolve(onRemove(removedFile)) // wrap in a promise (just in case)
```
dpikt: Shouldn't need to wrap in promise with `await`

src/forms/inputs/file-input/file-input.js
chawes13: So then the onChange in the file input would only call this?

dpikt: Yes, it would have no knowledge of `onChange` being overridden

src/forms/inputs/file-input/helpers.js
chawes13: I believe spread is necessary because `.files` is array-like (similar to arguments)

dpikt: 😵gotcha.

src/forms/inputs/file-input/file-input.js
chawes13: What's your recommendation for handling the remove use case? That's using the input's `onChange` method as well

chawes13:
```jsx
<FileInput
  input={{...input, onChange: async (files) => { input.onChange(files)}
  onRemove={ (files) => input.onChange(files) }
/>
```
`onRemove` then is a prop defaults to `this.removeFile` but can be completely overridden?

dpikt: Hmm... so I guess instead of overriding onChange this is really adding an extra read step for new files. That makes sense. Maybe a different name would alleviate confusion? I assumed `onRead` was a callback for triggering side effects but it's really part of the main flow... maybe `postProcess` or something? I'm probably overthinking it 😉Another option could be to overwrite `readFile` and pass that in:

```jsx
...
return (
 <FileInput
      input={ input }
      // called with all new files
      readFiles={ async (files) => {
          // same util used in fileinput
          const filesWithUrls = await readFileUrls(files)
         // upload them to cloudinary and return the result
          
      }}
```

chawes13: I like the latter approach!

chawes13:
@dpikt Take 3!

src/forms/inputs/cloudinary-file-input.js
```diff
@@ -41,37 +40,58 @@ import classnames from 'classnames'
...
+          })
+          
+          uploadedFiles = await Promise.all(uploadFilePromises)
+        } catch (e) {
+          onUploadFailure(e)
+          throw e
+        }
+        
+        onUploadSuccess(uploadedFiles)
+        return input.onChange(uploadedFiles)
```
dpikt: Can just return the files here right?

chawes13: 😳 😳 🤦‍♂

chawes13:
@dpikt In terms of thinking about the migration path, should we leave in the `onLoad` callback?

dpikt:
@chawes13 I'd say go with whatever seems like the best state for the library. Now's the time to make breaking changes if we want them! I wouldn't leave it in just for backwards compat, but if you can think of a valid use case for it then leave it in?

danparnella:
@chawes13 @dpikt Any chance this is close to being merged in? I mainly just need [this commit](https://github.com/LaunchPadLab/lp-components/pull/388/commits/e0d1ec331aed4e49643ab5d26a09c642aa41b6bf) to resolve an ADA compliance issue on artlook so I could also make a separate PR with just that?

dpikt:
@danparnella creating a separate PR would be the quickest path!

PR #389, opened 2019-10-15
Paginator focus styles

chawes13:
- Adds focus ring to paginator links


chawes13:
@marchwiany It could be worth an audit of all the places we're setting `outline: none` (and not replacing the `:focus` state in this library.

PR #397, opened 2019-11-14
Error label props defect

### LINKED ISSUE #396: Extra attributes on Input errors
Screen Shot 2019-11-14 at 10 03 19 AM

value and type are being added to the input's error message

chawes13:
Resolves #396 

**Notes**:
- `InputError` passes down additional props as it is often used as a standalone component for things like form-wide errors (i.e., `_error` on sign-in)
- A custom `errorComponent` should have access to all props, so that it can handle things like passing event handlers
- A standard `InputError` used in a `LabeledField` should not have all of the props passed down to it since the user cedes control over what gets spread to the `<span>`, hence making the distinction between the custom and standard error component in `LabeledField`


src/forms/labels/labeled-field.js
```diff
@@ -83,17 +83,18 @@ function LabeledField ({
...
   meta: { error, touched, invalid },
   className,
-  errorComponent: ErrorComponent = InputError,
+  errorComponent: ErrorComponent,
   labelComponent: LabelComponent = InputLabel,
   children,
   hideErrorLabel,
   ...rest
 }) {
+  const baseErrorProps = { error, invalid, touched, name }
```
dpikt: Maybe just pass these to the component directly rather than assigning it to a variable since `...rest` includes them anyway.

chawes13: @dpikt Since these values are destructured from props, aren't they not included in `...rest`?

dpikt: You're right, I got confused by the nesting. In that case we might want to pull them out a different way in case the custom component wants access to the other values stored in `input` and `meta`.

PR #409, opened 2020-01-27
Add masked input component

### LINKED ISSUE #44: Add MaskedInput component
So when we define our input, we could do something like this:
```
<Field 
   name="cardNumber"
   component={Input}
   mask="9999-9999-9999-9999"
/>
```

dpikt:
Resolves #44


dpikt:
@chawes13 I haven't. Just looking for a basic sanity check around entering a value?

chawes13:
@dpikt Yeah, exactly. I remember running into some issues with that with `react-maskedinput` so just want to make sure we're confident with the base case before adding it to the library

dpikt:
@chawes13 confirmed base case works with numeral mask

Screen Shot 2020-01-31 at 2 25 16 PM

PR #426, opened 2020-03-30
Add modal component

dpikt:
Adds a modal component that is untethered to the usage of redux / the HOC pattern. We've been using this along with local state on Prosci and it's been simple and intuitive so far (far less magical than the old method).

Ideally, I'd like to include this pattern rather than the HOC one as the recommended approach in the next version of the onboarding project (despite there being fewer examples in our projects). Let me know what you think about that idea!

Note: ideally `modal` from `lp-hoc` would render this component directly, however that's not possible at the moment since it would cause a circular dependency between the libs. This could be addressed later if it becomes an issue with making updates to one or both.

## Author Checklist

- [x] Add unit test(s)
- [ ] Update version in package.json
- [x] Assign dev reviewer


.size-limit.js
```diff
@@ -1,7 +1,7 @@
 module.exports = [
   {
     path: 'lib',
-    limit: '180 KB',
+    limit: '185 KB',
```
chawes13, line 4: 🙂

src/modal.js
```diff
@@ -0,0 +1,85 @@
...
+const defaultProps = {
+  hideCloseButton: false,
+}
+
+function getRootElement() {
+  // Skip in SSR mode
+  if (isServer()) return
+  // Note that this expects a root element with id "root"
+  // eslint-disable-next-line no-undef
+  return window.document.querySelector('#root')
```
chawes13, line 48: Is there a reason why we couldn't use `body` here?

src/modal.js
```diff
@@ -0,0 +1,85 @@
...
+      <div className="modal-content">{children}</div>
+      {!!onClose && !hideCloseButton && (
+        <>
+          <button
+            onClick={onClose}
+            className="modal-close"
+            aria-label="Close Modal"
+          >
+            ×
+          </button>
```
chawes13, line 75: Can you see a use case where it would be beneficial to allow a user to customize the close component?

dpikt: That could definitely be a future enhancement! Although they could also hide the default and place a new one in the modal body quite easily.

src/modal.js
dpikt: I think the only case would be if there were multiple apps mounted in the body... which is not our use case 😛 I can update!

dpikt: (this approach also wouldn't work for that case)

dpikt:
@chawes13 what do you think about including this approach in the onboarding project?

chawes13:
@dpikt love it

dpikt:
Resolved by #443

PR #432, opened 2020-07-07
Expand node support

chawes13:
The purpose of this PR is to expand node support to allow for using the latest version of the [AWS Connect library](https://github.com/amazon-connect/amazon-connect-streams) (`>=12.0.0`).

I don't think this is a breaking change, since `@size-limit/file@3.0.0` is only compatible with node `>=10.0.0`. Let me know if I should take this as an opportunity to bump to 4 **or** if you want to keep support for 8

## Author Checklist

- [x] ~Add unit test(s)~
- [x] ~Update documentation (if necessary)~
- [x] ~Add story to storybook (if necessary)~
- [x] Assign dev reviewer


dpikt:
Let's keep "support" for Node 8. Let's also add `^12` as an explicit version allowance to make this more future-proof.

chawes13:
@dpikt Done!

package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "@launchpadlab/lp-components",
-  "version": "3.31.3",
+  "version": "3.31.4",
   "engines": {
-    "node": "^8.0.0 || ^10.13.0"
+    "node": "^8.0.0 || ^10.13.0 || ^11.0.0 || ^12.0.0"
```
dpikt, line 5: Do we want to support 11 since it's not LTS? We don't support 7 or 9 😏

chawes13:
@dpikt I had to update tests to account for the sort algorithm changing in 11

package.json
chawes13: Doneso

chawes13:
Latest Travis report: https://travis-ci.org/github/LaunchPadLab/lp-components/builds/706629559

Issue #320, opened 2019-02-05
Add switch input component


dpikt:
We've got a simple one on apex, however this looks more robust: https://react-switch.netlify.com/

PR #433, opened 2020-07-08
Added <Switch> component, tests, and storybook

mwislek:
Resolves issue: https://github.com/LaunchPadLab/lp-components/issues/320 by adding a new `<Switch>` React component, based on https://react-switch.netlify.app.

## Author Checklist

- [x] Add unit test(s)
- [x] Assign dev reviewer


mwislek:
@dpikt . Let me know if you'd like to walk through this in a meeting and I can set something up.

dpikt:
@marchwiany what do you think about the default icons on this component? My instinct is to remove them but they might be necessary for a11y?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@launchpadlab/lp-components",
-  "version": "3.31.3",
+  "version": "3.31.4",
```
dpikt, line 3: Should be minor change since this is an enhancement

https://github.com/LaunchPadLab/opex-public/blob/master/gists/npm-package-guidelines.md#pull-requests-and-deployments

mwislek: @dpikt Cool. I added this link to the PR template as well. Let me know if there's a problem with that.

dpikt: Good idea!

marchwiany:
@dpikt yeah we should leave them since "color should not be used as the only visual means to convey information, indicate an action, or distinguish an element".

dpikt:
@mwislek this is good to merge to master 👍

mwislek:
Hi @dpikt . We're getting an interesting Travis error:
  Package size limit has exceeded by 795 B
258  Size limit: 185 KB
259  Size:       185.78 KB with all dependencies, minified and gzipped
260  
261  Try to reduce size or increase limit at .size-limit.js

Looks like all of our hard work (code) has pushed us over our size limits. Do you have any idea who/what the culprit might be? `yarn size` merely returns the same message with no details on individual components. Looking at individual file sizes wasn't too revealing either.

Also, you'll need to approve the PR in order for me to merge. Thanks!

dpikt:
Ah yeah, it's probably from adding `react-switch`. That should be fine (this library is modular anyway). That check is mostly to make sure we don't add anything huge by accident. You can bump the limit to 200kb to get it to pass 👍

dpikt (approved):

mwislek:
Thanks David! Finally merging!

Issue #435, opened 2020-07-16
Table column does not sort correctly when using `valueGetter`


dpikt:
Because it isn't passed to the underlying `sortable` instance. 

This bug could be a good opportunity to get rid of that underlying dependency. (see #414)

cc @mwislek

dpikt:
It seems like the component tests are pretty comprehensive so a refactor to hooks should be fairly safe. One exception is the `initialAscending` prop, so we should add a test for that as part of this fix.

PR #437, opened 2020-07-23
Refactor table sort

mwislek:
This PR resolves the issue [Table column does not sort correctly when using `valueGetter`](https://github.com/LaunchPadLab/lp-components/issues/435) by facilitating default sort when a `valueGetter` is specified. This PR also refactors the `SortableTable` component to use React hooks rather than the `sortable` HOC.

## Author Checklist

- [x] Add unit test(s)
- [x] Assign dev reviewer


mwislek:
@dpikt Not sure about the version number. This PR is a patch in the sense that it fixes a bug. But I did add tests (which were actually missing, not necessarily "new). Lemme know what you think.

dpikt:
Patch!

src/tables/helpers/apply-value-getters.js
```diff
@@ -0,0 +1,16 @@
+export default function applyValueGetters(columns, data) {
+  let annotatedData = data
+
+  columns.forEach(({ name, valueGetter }) => {
+    if (valueGetter) {
+      annotatedData = data.map((datum) => {
+        return {
+          ...datum,
```
dpikt, line 8: This will mess with the shape of the object when it's passed to a custom row or cell component. It would be better to just call `valueGetter()` twice, when sorting and when rendering.

src/tables/sortable-table.js
```diff
@@ -36,37 +36,89 @@ import classnames from 'classnames'
...
   rowComponent: Types.component,
   headerComponent: Types.component,
-  className: PropTypes.string,
-  ...sortablePropTypes,
 }
-
 const defaultProps = {
-  className: ''
+  className: '',
+  initialAscending: true,
```
dpikt, line 54: Let's add a default prop for every non-required prop, even if it's just `null`.

src/tables/sortable-table.js
```diff
@@ -36,37 +36,89 @@ import classnames from 'classnames'
...
-  sortPath,
-  setSortPath,
-  setSortFunc,
+  onChange,
   rowComponent,
   headerComponent,
-  className,
 }) {
-  const data = (controlled || disableSort) ? unsortedData : sort(unsortedData)
+  const [sortControls, setSortControls] = useState({
```
dpikt, line 75: I'd suggest splitting this into three `useStates`. It might seem more verbose at first but it will save you some potential headaches, and be more performant too.

src/tables/sortable-table.js
```diff
@@ -36,37 +36,89 @@ import classnames from 'classnames'
...
+
+    setSortControls({
+      ...sortControls,
+      ascending: newAscending,
+      sortPath: newSortPath,
+      sortFunc: newSortFunc,
+    })
+  }
+
+  useEffect(() => {
```
dpikt, line 101: Instead of using `setData` declaratively, you can derive the sorted data using `useMemo`.

```
const sortedData = useMemo(() => {...}, [ ascending, sortPath, sortFunc ])
```

And onChange can be triggered within handleColumnChange.

src/tables/sortable-table.js
```diff
@@ -109,43 +155,39 @@ function SortableTable ({
 SortableTable.propTypes = propTypes
 SortableTable.defaultProps = defaultProps
 
-const WrappedTable = sortable()(SortableTable)
-
-// Passes relevant sortable props
-function SortableTableWrapper ({ initialColumn, children, disableSort, disableReverse, onChange, ...rest }) {
+// Passes relevant SortableTable props and applies valueGetters...
+function SortableTableWrapper({
```
dpikt, line 159: Now that we're not using a HOC, you can remove this wrapper and move the functionality into the main component.

test/tables/sortable-table.test.js
```diff
@@ -202,11 +202,26 @@ test('initialColumn determines inital sortPath and sortFunc', () => {
   expect(mySort).toHaveBeenCalled()
 })
 
+test('initialColumn can be default sorted descending', () => {
```
dpikt, line 205: Also add a test case for sorting with valueGetter to make sure the bug is fixed!

src/tables/helpers/apply-value-getters.js
mwislek: Updated, as discussed

src/tables/sortable-table.js
mwislek: I did this except for the header/row component definitions. `null` is a valid component definition, which caused all of the tests to break. I could explicitly initialize to `undefined` but that would seem to default the purpose.

src/tables/sortable-table.js
mwislek: Done, as discussed

src/tables/sortable-table.js
mwislek: Done.

src/tables/sortable-table.js
mwislek: Done.

test/tables/sortable-table.test.js
mwislek: Done.

src/tables/sortable-table.js
dpikt: Ah, I guess that's what I get for using [destructuring defaults](https://zaiste.net/posts/javascript-destructuring-assignment-default-values/#:~:text=Default%20values%20in%20destructuring%20assignement,values%20in%20the%20destructuring%20statement.&text=You%20can%20combine%20default%20values%20with%20renaming%20in%20the%20destructuring%20assignment.) too heavily. Fine to leave it for now.

src/tables/sortable-table.js
```diff
@@ -36,37 +36,118 @@ import classnames from 'classnames'
...
+    initialSortPath: '',
+    initialSortFunc: null,
+    initialValueGetter: null,
+  }
+  if (!initialColumn) return defaultControls
+
+  const initialProps = columns.filter(col => col.name === initialColumn).pop()
+  // Exceptional situation-- an initial column was specified but no column data
+  // exists for the named column...
+  if (!initialProps) throw new TableColumnError('initial column has no column definition')
```
dpikt, line 73: Sorry, I was misleading when I said throw a custom error- the standard error class is fine for this, as long as the message is helpful (which it is). Catching certain error instances is pretty rare in JS, so unless you need your error to do something really special, subclassing is not necessary.

src/tables/sortable-table.js
```diff
@@ -36,37 +36,118 @@ import classnames from 'classnames'
...
 
-function SortableTable ({
-  columns,
+function getInitialSortControls(initialColumn, columns) {
+  const defaultControls = {
+    initialSortPath: '',
+    initialSortFunc: null,
+    initialValueGetter: null,
+  }
+  if (!initialColumn) return defaultControls
```
dpikt, line 68: You can define this as a constant outside of the function.

src/tables/sortable-table.js
```diff
@@ -36,37 +36,118 @@ import classnames from 'classnames'
...
+    if (sortFunc) {
+      const sorted = [...unsortedData].sort(sortFunc)
+      if (!ascending && !disableReverse) sorted.reverse()
+      return sorted
+    }
+    else {
+      const order = ascending ? 'asc' : 'desc'
+      const sorted = orderBy(
+        unsortedData,
+        (item) => valueGetter ? valueGetter(item) : get(sortPath, item),
```
dpikt, line 119: Since this now uses `valueGetter`, you should add that to the dependencies array of `useMemo`.

src/tables/sortable-table.js
mwislek: Subclassing, to me, really provides nothing more than documentation. Since it wasn't done elsewhere in this library, I removed the subclass.

src/tables/sortable-table.js
mwislek: Done

src/tables/sortable-table.js
mwislek: Good one. Fixed.

dpikt:
@mwislek Ok, travis is back, merge at will!

PR #443, opened 2020-08-31
Add modal component to V3

dpikt:
A redo of #426 that doesn't rely on the unmerged `style` branch. Since this is a new component, requiring a style change to use it is not technically breaking 🙂 

## Author Checklist

- [x] Add unit test(s)
- [x] Assign dev reviewer


src/modal.js
```diff
@@ -0,0 +1,93 @@
...
+const defaultProps = {
+  hideCloseButton: false,
+}
+
+function getRootElement() {
+  // Skip in SSR mode
+  if (isServer()) return
+  // Note that this expects a root element with id "root"
+  // eslint-disable-next-line no-undef
+  return window.document.querySelector('#root')
```
chawes13, line 56: Not sure if we discussed this question on the old PR or something else, but any concerns with using `body` here?

dpikt: Deja vu! I must've forgotten to copy that commit over. https://github.com/LaunchPadLab/lp-components/pull/426#discussion_r402605702

src/modal.js
```diff
@@ -0,0 +1,93 @@
...
+function Modal({ onClose, hideCloseButton, children, ...rest }) {
+  return (
+    <ReactModal
+      isOpen
+      onRequestClose={onClose}
+      portalClassName="modal"
+      className="modal-inner"
+      overlayClassName="modal-fade-screen"
+      bodyOpenClassName="modal-open"
+      appElement={getRootElement()}
```
chawes13, line 69: Should this be memoized as a ref?

dpikt: Good question- I think the function is basic enough that it's not worth the overhead.

PR #444, opened 2020-09-01
Bump react peer dependency to 16.8

dpikt:
Pivoting on the strategy for v4, to start making more incremental progress. _Technically_ we already started using hooks without this change (my mistake) but this formalizes the upgrade.

## Author Checklist

- [x] Assign dev reviewer


chawes13 (approved):
🎣

PR #446, opened 2020-09-28
Remove `FlashMessage` dependency on `redux-flash`

### LINKED ISSUE #415: Remove dependency on `redux-flash`
This dependency is only used for a proptype shape, which could be recreated in-library or swapped for a better abstraction.

dpikt:
Resolves #415. `FlashMessageContainer` has been left as-is for backwards compatibility with the current client template.


migration-guides/v5.0.0.md
```diff
@@ -0,0 +1,17 @@
...
+
+(Note: this change is not necessary if you are using `FlashMessageContainer` instead of `FlashMessage` directly.)
+
+```jsx
+
+// Before, where "message" was a redux-flash message object:
+<FlashMessage message={message} />
+
+// After:
+<FlashMessage isError={message.isError} {message.props}>{message.message}</FlashMessage>
```
chawes13, line 15: Can you explain what `{message.props}` is accomplishing here?

src/indicators/flash-message.js
```diff
@@ -1,51 +1,59 @@
...
- * A component that displays a flash message generated by [redux-flash](https://github.com/LaunchPadLab/redux-flash).
- * Any message props will be passed through to this component.
- * 
+ * A component that displays a flash message.
+ *
  * @name FlashMessage
  * @type Function
- * @param {Object} message - The flash message that will be displayed.
+ * @param {String} children - The flash message that will be displayed.
+ * @param {Boolean} [isError] - A flag to indicate whether the message is an error message.
```
chawes13, line 12: Can you specify the default here?

src/indicators/flash-message.js
```diff
@@ -1,51 +1,59 @@
...
+      {...rest}
+    >
+      {onDismiss && (
+        <button type="button" className="dismiss" onClick={() => onDismiss()}>
           ×
         </button>
-      }
-      <p> { message.message } </p>
+      )}
+      <p> {children} </p>
```
chawes13, line 56: Now that we're accepting children, it's not necessarily clear to me that it can / should only accept phrasing content. In other words, should children be rendered in a `p` tag?

dpikt: Yeah, I thought about that. I decided to leave it in for backwards compat, but I can move it to `FlashMessageContainer` instead.

migration-guides/v5.0.0.md
dpikt: That's a typo- let me know if it's still unclear.

chawes13: I could not wrap my brain around that haha. Makes sense now!

chawes13 (approved):
👍

PR #452, opened 2020-10-19
Modify file input markup to allow for focus styling

danparnella:
Enables focus styling on the `span` (when focusing on the `input`) by moving it after the `input`, which allows for the use of adjacent sibling css selector (`+`).

## Author Checklist

- [x] Add unit test(s)
    - I attempted to add one to check for a style change, but doesn't look like there's a good way to grab non-inline styles applied by external css files. Open to any ideas on how to apply a test here!
    - Would you consider this a patch or minor update?
    - Should anything be updated to mention adding adjacent sibling focus styles? I haven't seen that kind of documentation mentioned for other components.
- [x] Assign dev reviewer

### DIFF
--- a/src/forms/inputs/file-input/file-input.js
+++ b/src/forms/inputs/file-input/file-input.js
@@ -5,6 +5,7 @@ import { LabeledField } from '../../labels'
 import FilePreview from './file-preview'
 import ImagePreview from './image-preview'
 import { noop, generateInputErrorId, isServer } from '../../../utils'
+import classnames from 'classnames'
 
 /**
  *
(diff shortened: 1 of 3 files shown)


src/forms/inputs/file-input/file-input.js
```diff
@@ -114,22 +115,29 @@ class FileInput extends React.Component {
...
+                accept,
+                'aria-labelledby': name + '-label',
+                'aria-describedby': hasInputError(meta)
+                  ? generateInputErrorId(name)
+                  : null,
+              }}
+            />
+            {/* Include after input to allowing for styling with adjacent sibling selector */}
+            <span
+              className={classnames('fileupload-exists', wrapperClass)}
```
bhennes2, line 135: Taking this `className` at face value, wouldn't we want `fileupload-exists` to be dynamic based upon the state of the component?

bhennes2 (commented):
Point of clarification for me on dynamic nature of the class names.

I think the versioning question comes back to is this a potential breaking change. Seems like stylistically it could be so I think a `MINOR` version (and potentially `MAJOR`) change would be necessary. Would defer to @dpikt on the mechanics/hygiene around that though.

src/forms/inputs/file-input/file-input.js
danparnella: Hmm, yeah I'm not sure the reasoning behind that naming since it should always be there (following the same logic that was there originally). Maybe it should be renamed to `fileupload-button`?

dpikt (approved):
Can you add a note about the conversion to a non-breaking change?

danparnella:
I was able to convert this to a non-breaking change by keeping the `wrapperClass` on the wrapping `div` and just making additional style modifications to achieve what we're looking for (a highlight outline around the whole file input "button"). If no style changes are made on current projects, there will be no breaking change to the component.

Current projects can even just add focus styles to the, now adjacent, `span` (without modifying any other styles) to at lease provide an outline around the text of the "button".

Otherwise, the "button" specific styles can be moved from the `wrapperClass` to the inner `span` to achieve the ideal state.

Issue #419, opened 2020-02-11
Make options and value required in TabBar


chawes13:
`v3.31.0` introduced an enhancement where the styling will be applied correctly to the first option (if any exist), but since this is a controlled input that doesn't really address the need in terms of functionality. a `Tablist` should always have a tab visible, so we should require that value is set. Related to that, options should be required as well because you can't really have a `value` set if there are no options

chawes13:
Note: this would be a breaking change, but could go in v4

PR #461, opened 2020-11-21
Made options & value as required for the TabBar component

AngelMDev:
Addresses: #419. 

This one I admit is a bit outside the area of my expertise, but I think these changes do the trick, assuming chaining `isRequired` works as I expect it to do. I guess one question that I have is if the linter is supposed to catch these, or if an error is supposed to print in dev console? I certainly saw none of these for both of the params, while other components do show a linter error. Lmk if I needed to do anything else here for that to show.

Bumping to a new major version.


chawes13 (changes requested):
Thanks for the PR! There are a few more subtle changes required here as a result of this change. Let me know if you have any questions or want to pair!

- Documentation should be updated to no longer mark options and values as optional (https://share.getcloudapp.com/Z4uqZ1k4) (the `[]` brackets denote optional)
- A migration guide should be created to help people understand what they need to do to upgrade from `5.0.0` to `6.0.0` ([migration guide example](https://share.getcloudapp.com/Z4uqZ1k4))
- Since the values are required, you no longer should have default props declared (https://share.getcloudapp.com/bLu0Jxwm)
- There is no longer the concept of "falling back" to the value of the first option, so the concept of [activeValue](https://share.getcloudapp.com/eDuwol8o) should be removed from the component
- Tests and StoryBook will need to be updated to account for these changes (e.g., https://github.com/LaunchPadLab/lp-components/blob/0ea9404056f8bece2fd3fe84e09de57222a7c985/test/controls/tab-bar.test.js#L71)

AngelMDev:
> @AngelMDev Do you have any bandwidth to make these adjustments?

Sorry, this has been off my radar for a really long time! Yes I can address this. On the upside, I now have a better idea of what I'm doing hah.

chawes13 (approved):
🎉  Nice work!

Issue #307, opened 2018-12-04
Pass all Table props down to rowComponent 


dpikt:

mwislek:
Hey @dpikt ! 
When you say "all Table props" do you really mean "every single one"? I have a feeling you mean "a few". Here's the current list of table props:
```
function SortableTable({
  className,
  children,
  data: unsortedData,
  initialAscending,
  initialColumn,
  disableReverse,
  disableSort,
  controlled,
  onChange,
  rowComponent,
  headerComponent,
}) {
```

...and here's what's currently passed to the `rowComponent`:
```
function TableRow ({
  columns,
  rowComponent: RowComponent = DefaultRowComponent,
  rowData,
}) {
  return (
    <RowComponent { ...{ data: rowData } }>
```

I could see passing down `className`. I have another PR (https://github.com/LaunchPadLab/lp-components/pull/462) that allows one to pass arbitrary props to a table, all of which could also be passed to the `rowComponent`. Were you thinking of something else? Thanks!

dpikt:
Hmm, honestly I don't remember the exact use case for this, I created it two years ago and didn't exactly clarify in the description 🤦 I think the idea was actually less about props and more about state (which were passed as props from a HOC at the time) which can be passed down manually when you define the component. That is, if you want a rowComponent that changes how it renders based on `sortPath`, right now it has no access to that information. So I think passing the internal table state as props to the row component (`ascending, sortPath, sortFunc, valueGetter`) would be adequate for this task.

mwislek:
Thanks David! Your approach makes sense and will be implemented as such.

chawes13:
Resolved by https://github.com/LaunchPadLab/lp-components/pull/464

PR #464, opened 2021-01-08
Row component props

mwislek:
Resolves issue: https://github.com/LaunchPadLab/lp-components/issues/307

This PR provides the following table state values to the user-defined `RowComponent`:

- ascending
- sortPath
- sortFunc
- valueGetter

Developers may use these additional state values to enhance the functionality of their customer `RowComponent`.

## Author Checklist

- [x] Add unit test(s)
- [x] Assign dev reviewer

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@launchpadlab/lp-components",
-  "version": "5.1.1",
+  "version": "5.3.0",
   "engines": {
     "node": "^8.0.0 || ^10.13.0 || ^12.0.0"
   },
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@launchpadlab/lp-components",
-  "version": "5.1.1",
+  "version": "5.3.0",
```
mwislek, line 3: Let me know if you think this should be a major version update instead. While we did change an existing interface, the change only added to the interface so the change is completely backwards compatible.

dpikt (approved):
Looks great!

package.json
dpikt: Backwards compatible -> minor change is good!

Issue #470, opened 2021-02-11
Cannot programmatically focus the textarea input in the Textarea component


danparnella:
Without using DOM traversal with standard Javascript, I haven't been able to find a way to programmatically focus the textarea input in the `Textarea` component. My suggestion is to allow for a `forwardRef` prop on `Textarea` that would be passed down to the textarea input. That ref could then be used to target the input itself for focusing.

chawes13:
I ran into this issue yesterday with our base `Input` component.

Can we use [`React.forwardRef`](https://reactjs.org/docs/forwarding-refs.html)? Here's a reference from Styled Components: https://github.com/system-ui/theme-ui/pull/817/files

Ideally this would be applied to all of our inputs, but I'm ok with that being a separate PR to expedite the fix for Artlook

chawes13:
That being said, we will probably run into an issue with how `redux-form`s `Field` component wraps our components. From the [documentation](https://redux-form.com/8.3.0/docs/api/field.md/#-code-forwardref-boolean-code-optional-), it looks like their ref forwarding implementation via `getRenderedComponent` only works if the component is a class component.

I initially thought we might be able to get away with something like the below, but haven't tested it. I did see one SO post that said it wasn't working for them.
```jsx
<Field
  name="content"
  component={(props) => <Input ref={inputRef} {...props} />}
/>

danparnella:
Yeah, that was originally was my hope (to just export our component with `forwardRef`), but yes I ran straight into that class restriction ☹️ . Your last suggestion is basically what I ended up doing in my band-aid override, but without the fallback to using the correct `forwardRef` implementation (and I think I could use `forwardRef` in my `Field` in artlook because we're using an older version of `redux-form`), so I think what you have is the way to go.

I plan on getting a PR up for this today!

danparnella:
@chawes13 I'm running into some bumps in the road represented by [this issue](https://github.com/facebook/react/issues/16653). Does that make sense in that I need to define prop types and default props _after_ exporting `Textarea` as a `forwardRef`?

chawes13:
@danparnella Did you try the solution outlined in this comment? https://github.com/facebook/react/issues/16653#issuecomment-564423981

danparnella:
Got it, got it. Ok, just wanted to make sure I didn't go too off the rails with our patterns 👍 .

chawes13:
Uncharted territory 🤪

PR #472, opened 2021-02-12
Textarea Ref Forwarding

danparnella:
Resolved Issue: https://github.com/LaunchPadLab/lp-components/issues/470
Description: Allows for a ref to be provided (through a new `forwardedRef` prop) on `Textarea` that will be passed down to the inner `textarea` input.

## Author Checklist

- [x] Add unit test(s)
- [x] Assign dev reviewer

### DIFF
--- a/src/forms/inputs/textarea.js
+++ b/src/forms/inputs/textarea.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, {forwardRef} from 'react'
 import PropTypes from 'prop-types'
 import classnames from 'classnames'
 import { blurDirty, fieldPropTypes, hasInputError, omitLabelProps } from '../helpers'
@@ -7,7 +7,8 @@ import { compose, filterInvalidDOMProps, generateInputErrorId } from '../../util
 
 /**
  *
- * A textarea input that can be used in a `redux-forms`-controlled form. Optionally displays a character count.
+ * A textarea input that can be used in a `redux-forms`-controlled form.
+ * Forwards ref down to textarea input and optionally displays a character count.
  *
  * @name Textarea
  * @type Function
@@ -15,6 +16,7 @@ import { compose, filterInvalidDOMProps, generateInputErrorId } from '../../util
  * @param {Object} meta - A `redux-forms` [meta](http://redux-form.com/6.5.0/docs/api/Field.md/#meta-props) object
  * @param {Number} [maxLength] - The maximum allowed length of the input
  * @param {Boolean} [hideCharacterCount=false] - Whether to hide the character count if given a maxLength
+ * @param {Ref} [forwardedRef] - A ref to be forwarded to `textarea` input when used with `redux-forms`
  * @example
  *
  * function BiographyForm ({ handleSubmit, pristine, invalid, submitting }) {
@@ -36,21 +38,29 @@ import { compose, filterInvalidDOMProps, generateInputErrorId } from '../../util
 const propTypes = {
   ...fieldPropTypes,
   hideCharacterCount: PropTypes.bool,
-  maxLength: PropTypes.oneOfType([PropTypes.number, PropTypes.bool])
+  maxLength: PropTypes.oneOfType([PropTypes.number, PropTypes.bool]),
+  forwardedRef: PropTypes.oneOfType([
+    PropTypes.func,
+    PropTypes.shape({
+      current: PropTypes.instanceOf(Element), // eslint-disable-line no-undef
+    }),
+  ]),
 }
 
 const defaultProps = {
   maxLength: null,
   hideCharacterCount: false,
+  forwardedRef: null,
 }
 
-function Textarea (props) {
+const Textarea = forwardRef(function Textarea (props, ref) {
   const {
     input: { name, value, onBlur, onChange },
     meta, // eslint-disable-line no-unused-vars
     hideCharacterCount,
     className,
     maxLength,
+    forwardedRef,
     ...rest
   } = omitLabelProps(props)
   return (
@@ -72,16 +82,16 @@ function Textarea (props) {
           value,
           onBlur,
           onChange,
+          ref: forwardedRef || ref,
           'aria-describedby': hasInputError(meta) ? generateInputErrorId(name) : null,
           ...filterInvalidDOMProps(rest),
         }}
       />
     </LabeledField>
   )
-}
+})
 
 Textarea.propTypes = propTypes
-
 Textarea.defaultProps = defaultProps
 
 export default compose(
--- a/src/forms/inputs/textarea.js
+++ b/src/forms/inputs/textarea.js
@@ -1,4 +1,4 @@
-import React, {forwardRef} from 'react'
+import React, { forwardRef } from 'react'
 import PropTypes from 'prop-types'
 import classnames from 'classnames'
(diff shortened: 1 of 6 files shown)


src/forms/inputs/textarea.js
```diff
@@ -7,14 +7,16 @@ import { compose, filterInvalidDOMProps, generateInputErrorId } from '../../util
...
+ * A textarea input that can be used in a `redux-forms`-controlled form.
+ * Forwards ref down to textarea input and optionally displays a character count.
  *
  * @name Textarea
  * @type Function
  * @param {Object} input - A `redux-forms` [input](http://redux-form.com/6.5.0/docs/api/Field.md/#input-props) object
  * @param {Object} meta - A `redux-forms` [meta](http://redux-form.com/6.5.0/docs/api/Field.md/#meta-props) object
  * @param {Number} [maxLength] - The maximum allowed length of the input
  * @param {Boolean} [hideCharacterCount=false] - Whether to hide the character count if given a maxLength
+ * @param {Ref} [forwardedRef] - A ref to be forwarded to `textarea` input when used with `redux-forms`
```
bhennes2, line 19: Perhaps I'm misunderstanding but since we are using `forwardRef` imported from `React`, is using this prop actually coupled to `redux-forms`?

chawes13: We're adding this to help support usage in redux-forms, but technically it will always accept a `forwardedRef` prop, regardless of the form library. I think I'd prefer for this comment to be agnostic of the form library.

src/forms/inputs/textarea.js
```diff
@@ -36,21 +38,29 @@ import { compose, filterInvalidDOMProps, generateInputErrorId } from '../../util
 const propTypes = {
   ...fieldPropTypes,
   hideCharacterCount: PropTypes.bool,
-  maxLength: PropTypes.oneOfType([PropTypes.number, PropTypes.bool])
+  maxLength: PropTypes.oneOfType([PropTypes.number, PropTypes.bool]),
+  forwardedRef: PropTypes.oneOfType([
+    PropTypes.func,
+    PropTypes.shape({
+      current: PropTypes.instanceOf(Element), // eslint-disable-line no-undef
```
chawes13, line 45: Typically it's safer to use `// eslint-disable-next-line <rule>` on the line above since eslint and/or prettier can potentially wrap the line when making automated changes

stories/forms/inputs/textarea-story.js
```diff
@@ -36,3 +38,10 @@ storiesOf('Textarea', module)
       hideCharacterCount={true}
     />
   ))
+  .add('with a forwardedRef', () => (
+    <Textarea
+      input={inputProps}
+      meta={{}}
+      forwardedRef={inputRef}
+    />
+  ))
```
chawes13, line 47: Because this story is using the input directly, we should be able to add a story for a "proper" forwarded ref as well (i.e., `ref={inputRef}`)

test/forms/inputs/textarea.test.js
```diff
@@ -72,9 +72,24 @@ test('Textrea does not receive invalid dom attributes', () => {
...
-    onClickLabel: () => 'foo'
+    onClickLabel: () => 'foo',
   }
-  
+
   const wrapper = mount(<Textarea {...props} />)
   expect(wrapper.find('textarea').prop('onClickLabel')).toBe(undefined)
 })
+
+test('Textarea passes down forwardedRef to input correctly', () => {
```
chawes13, line 82: Similarly to the story comment, I'd like to see a test for `ref` getting passed directly to TextArea

src/forms/inputs/textarea.js
```diff
@@ -36,21 +38,29 @@ import { compose, filterInvalidDOMProps, generateInputErrorId } from '../../util
...
 }
 
 const defaultProps = {
   maxLength: null,
   hideCharacterCount: false,
+  forwardedRef: null,
 }
 
-function Textarea (props) {
+const Textarea = forwardRef(function Textarea (props, ref) {
```
chawes13, line 56: Can you include a screenshot of how this component displays in React dev tools? This approach is new to me and I want to make sure that the advice from that GitHub issue works!

chawes13:
@danparnella Sorry for the delay!

src/forms/inputs/textarea.js
chawes13: I realize that this seems directly contradictory with the precedent set with `meta` below, but I stand by it! 😆 😅

src/forms/inputs/textarea.js
danparnella: 👍 I attempted to explain more generally why you would use `forwardedRef`.

src/forms/inputs/textarea.js
danparnella: Looks like `meta` is actually being used now anyway 😆.

test/forms/inputs/textarea.test.js
danparnella: Good thing you called this out! A standard `ref` was **not** working correctly. The issue is we're wrapping `Textarea` in another component when composing with `blurDirty`. I ended up exporting that using `forwardRef` as well to pass the `ref` down to `Textarea`. Do you think that requires some additional new documentation on `blurDirty` to call that out? Any other ideas that might be cleaner?

src/forms/inputs/textarea.js
danparnella: Here it is after rendering it with a standard `ref` passed in: https://share.getcloudapp.com/YEuXONjN

chawes13: 🙌  that looks great

test/forms/inputs/textarea.test.js
chawes13: Oof 😬 . The challenge with HOCs is that it looks like for the `ref=` scenario to work, every single HOC would need to have `forwardRef` applied to it. It looks like the [React team's recommendation](https://reactjs.org/docs/forwarding-refs.html#forwarding-refs-in-higher-order-components) is to just use a custom prop name (the same name we choose actually) to pass down to the wrapped component.

Since we do have components that use multiple HOCs (e.g., `checkbox` uses `blurDirty` **and** `replaceEmptyStringValue`), I think it would actually be better to only accept the custom `forwardedRef` prop in `Textarea` (i.e., remove the `forwardRef()` usages, stories, and tests).

That being said, in the long run we could create a `withForwardedRef` HOC to apply to other inputs that would pass the custom ref prop down automatically (example below). There might also be other alternatives that we could explore by taking a look at some more established OS react component libraries. I would definitely consider all of that out of scope for this PR. As an aside, it always looks like the React team recommends the inclusion of `forwardRef` to warrant a major version change as it should be considered a breaking change (https://reactjs.org/docs/forwarding-refs.html#note-for-component-library-maintainers).

```js
function withForwardedRef(Component) {
  return React.forwardRef((props, ref) => <Component {...props} forwardedRef={ref} />)
}
```
Sorry for the reversal on this decision, but hopefully you found this exploration informative!

danparnella: All good, it was helpful to walk through how it would with HOCs. Ready for another look!

chawes13 (approved):

PR #473, opened 2021-02-18
Upgrade to stack version 20

danparnella:
Heroku builds no longer succeed:
```
This app is using the Cedar-14 stack, which has been end-of-life since May 1st 2019. The building of Cedar-14 apps is no longer supported as of November 2nd 2020.
```
Setting the latest stack (version 20 currently) will need to be completed on the Production app in Heroku, but the change made here will allow us to test that nothing breaks using the Review App (this change will also need to be merged in to `ensure that your Review Apps and Heroku CI runs use the same stack`).

Details here: https://devcenter.heroku.com/articles/upgrading-to-the-latest-stack

## Author Checklist

- [ ] Add unit test(s)
- [x] Assign dev reviewer

### DIFF
--- a/app.json
+++ b/app.json
@@ -23,5 +23,6 @@
     {
       "url": "heroku/nodejs"
     }
-  ]
+  ],
+  "stack": "heroku-20"
 }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@launchpadlab/lp-components",
-  "version": "5.4.0",
+  "version": "5.4.1",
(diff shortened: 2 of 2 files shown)


danparnella:
@chawes13 This appears to be working. I saw no issues after going through the entire storybook in the review app. What version bump would you consider this? It's not a breaking change and I don't think it requires any new documentation or tests,  so just a ^0.0.1?

chawes13 (approved):
Thank you 🙏  🙏

PR #493, opened 2021-10-18
Add className to spinner component

ghrlee:
quick change to the Spinner component to use the className "spinner" in place of id

## Author Checklist

- [x] Add unit test(s)
- [x] Assign dev reviewer


src/indicators/spinner.js
```diff
@@ -31,7 +31,7 @@ const propTypes = {}
 const defaultProps = {}
 
 function Spinner (props) {
-  return <div id="spinner" { ...filterInvalidDOMProps(props) } />
+  return <div className="spinner" { ...filterInvalidDOMProps(props) } />
```
chawes13, line 34: I think we should allow for the user to also specify an additional classname. We can do this using the `classnames` package. You can reference an example of this in `src/forms/buttons/button-area.js`.

migration-guides/v6.0.0.md
```diff
@@ -5,6 +5,7 @@ This version contains the following breaking changes:
 1. The `style` prop on `<Button />` and `<SubmitButton />` has been renamed to `variant`
 1. Labels for `<CheckboxGroup />` and `<RadioGroup />` now use a `<legend>` html tag rather than a `<label>`
 1. The `hideLabel` prop on `<RangeInput />` has been renamed to `hideRangeValue`
+1. The tag on `<Spinner />` now uses the classname `spinner` in place of an id
```
chawes13, line 8: `className` is a JSX specific prop, so I think you can just use "class" here!

migration-guides/v6.0.0.md
```diff
@@ -77,3 +78,14 @@ Any `<RangeInput />` with a `hideLabel` prop should be updated to use `hideRange
 // After
 <RangeInput input={rangeInputProps} hideRangeValue={true} />
 ```
+#### 6. The tag on `<Spinner />` now uses the classname `spinner` in place of an id
```
chawes13, line 81: Can you merge in the latest from v6 into your branch? 6, 7, and 8 were recently added

migration-guides/v6.0.0.md
```diff
@@ -77,3 +78,14 @@ Any `<RangeInput />` with a `hideLabel` prop should be updated to use `hideRange
...
+
+`<Spinner />` utilizes a classname `spinner` in order to accomodate views where more than one spinners are necessary.
+
+```jsx
+// Before
+<div id="spinner" { ...filterInvalidDOMProps(props) } />
+
+// After
+<div className={classnames('spinner', props.className)} />
```
chawes13, line 90: This guide is focused on the change that a developer has to make to their existing code, not the actual code change that was made. The breaking change here is the styling change. If a dev updated their project to use this version of `lp-components`, the actually spinner animation wouldn't show up until they modified their scss files (i.e., replacing any reference of `#spinner` with `.spinner`)

migration-guides/v6.0.0.md
```diff
@@ -9,6 +9,7 @@ This version contains the following breaking changes:
 1. `<CloudinaryFileInput />` and `<FileInput />` components now store an array of objects representing the file data as opposed to a string
 1. `<CloudinaryFileInput />` and `<FileInput />` components no longer accept an `onLoad` prop
 1. The `previewComponent` for a file input no longer receives a `value` prop and `file` is a file object (with the url)
+1. The tag on `<Spinner />` now uses the class `spinner` in place of an id which can be custom
```
chawes13, line 12: "...which can be custom" implies to me that the spinner class can be overridden. Our implementation allows for appending _additional_ classes, but not customizing the base class. Would you mind updating this language?

migration-guides/v6.0.0.md
```diff
@@ -197,12 +197,21 @@ const PreviewComponent = ({ file }) => {
 
 #### 9. The tag on `<Spinner />` now uses the classname `spinner` in place of an id
 
-`<Spinner />` utilizes a classname that can be custom in order to accomodate views where more than one spinners are necessary.
+`<Spinner />` utilizes a classname and can accept additional custom classes in order to accomodate views where more than one spinners are necessary.
```
chawes13, line 200: Typo: `accommodate`

migration-guides/v6.0.0.md
```diff
@@ -197,12 +197,21 @@ const PreviewComponent = ({ file }) => {
...
+  text-align: center;
+}
 
 // After
-<div className="custom" { ...filterInvalidDOMProps(props) } />
+.spinner {
+  text-align: center;
+}
+
+//
```
chawes13, line 213: Were you going to add a comment here?

src/indicators/spinner.js
```diff
@@ -26,12 +27,15 @@ import { filterInvalidDOMProps } from '../utils'
  *
  */
 
-const propTypes = {}
+const propTypes = {
+  className: PropTypes.string
```
chawes13, line 31: When a prop isn't required, it's best to declare your intent with a default prop. For this use case, an empty string would suffice.

chawes13 (commented):
Looking really good!! Just a few finishing touches and we'll be ready to merge 🤩

ghrlee:
@chawes13 sorry for the huge delay

migration-guides/v6.0.0.md
```diff
@@ -193,3 +194,24 @@ const PreviewComponent = ({ file }) => {
...
+}
+
+// After
+.spinner {
+  text-align: center;
+}
+
+// After
+.custom-spinner {
+  text-align: center;
```
chawes13, line 215: It'd be great if you chose another declaration here to really showcase the benefit of customizing just a specific spinner. When I read this example, you're not changing anything from above so it reads a little confusing to me. E.g., 
```css
.custom-spinner {
  margin-bottom: 10px;
}

File: migration-guides/v6.0.0.md [hunk 10]
```diff
@@ -193,3 +194,24 @@ const PreviewComponent = ({ file }) => {
...
+#spinner {
+  text-align: center;
+}
+
+// After
+.spinner {
+  text-align: center;
+}
+
+// After
```
chawes13, line 213: I don't think you need this here, since you have it on line 208

src/indicators/spinner.js
```diff
@@ -26,12 +27,15 @@ import { filterInvalidDOMProps } from '../utils'
  *
  */
 
-const propTypes = {}
+const propTypes = {
+  className: ''
```
chawes13, line 31: This isn't accurate. In `propTypes` you use `PropTypes` to declare the type of the prop. In `defaultProps`, you declare the default value for the properties that aren't required in `propTypes`

chawes13: Here's an example: https://github.com/LaunchPadLab/lp-components/blob/ead699069fa1a2c322c7a117b3c2c6a196c2b1c1/src/forms/inputs/input.js#L37-L45

test/indicators/spinner.test.js
```diff
@@ -3,23 +3,30 @@ import { mount } from 'enzyme'
...
   )
-  expect(wrapper.find('div#spinner').exists()).toBe(true)
+  expect(wrapper.find('div.spinner').exists()).toBe(true)
+})
+
+test('Spinner can have custom className', () => {
+  const wrapper = mount(
+    <Spinner className="custom" />
+  )
+  expect(wrapper.find('div.custom').exists()).toBe(true)
```
chawes13, line 17: Can you also assert that the `.spinner` class is present as well?

chawes13 (changes requested):
So close! Just a few finishing touches

src/indicators/spinner.js
```diff
@@ -28,10 +28,12 @@ import classnames from 'classnames'
  */
 
 const propTypes = {
-  className: ''
+  className: PropTypes.string,
 }
 
-const defaultProps = {}
+const defaultProps = {
+  className: 'spinner'
```
chawes13, line 35: If `className` isn't provided, your div will look like this:
```jsx
<div class="spinner spinner">
```
An empty string would be a more appropriate default (ignored by `classnames` util)

migration-guides/v6.0.0.md
```diff
@@ -193,3 +194,19 @@ const PreviewComponent = ({ file }) => {
...
+```scss
+// Before
+#spinner {
+  text-align: center;
+}
+
+// After
+.custom-spinner {
+  margin-bottom: 10px;
+}
```
chawes13, line 211: I think there might have been a miscommunication. I did not intend for you to also delete the code with `.spinner`. I think you need to show both.

chawes13 (changes requested):

chawes13 (approved):
🚀 nice!

Issue #458, opened 2020-11-03
Ability pass other props or rest to Cell Component


inveterateliterate:
I wanted to be able to pass some other props to my cell without creating a custom cell component, namely I just wanted to add a cypress tag but the cell just renders the value:

```
<Column
    name="dateCompleted"
    data-cy="dateCompleted"
/>
```

Right now the default cell component only takes a few defined props: https://github.com/LaunchPadLab/lp-components/blob/master/src/tables/components/table-row.js#L13

Other things we might want to pass could be some accessibility tags as well, such as aria-label, etc, so might be worthwhile to have the component take everything passed to it.

PR #503, opened 2021-11-10
allow cell to take custom properties like aria and data-cy per cell b…

aojin:
…y column

[Ability pass other props or rest to Cell Component #458](https://github.com/LaunchPadLab/lp-components/issues/458)


stories/tables/sortable-table.story.js
```diff
@@ -48,7 +48,7 @@ function compareCustomValue(a, b) {
 storiesOf('SortableTable', module)
   .add('default', () => (
     <SortableTable data={tableData}>
-      <Column name="name" />
+      <Column name="name"/>
```
chawes13, line 51: This change doesn't seem necessary

src/tables/components/table-row.js
```diff
@@ -14,8 +14,8 @@ const propTypes = {
 }
 
 const DefaultRowComponent = ({ children }) => <tr>{ children }</tr> // eslint-disable-line
-const DefaultCellComponent = ({ className, onClick, placeholder, value }) => // eslint-disable-line
-  <td { ...{ className, onClick }}>{ isNil(value) ? placeholder : value }</td>
+const DefaultCellComponent = ({ className, onClick, placeholder, value, ...rest }) => // eslint-disable-line
+  <td { ...{ className, onClick, ...rest }}>{ isNil(value) ? placeholder : value }</td>
```
chawes13, line 18: We should leverage the `filterInvalidDOMProps` util here. Otherwise, we'll see warnings like this:
Image 2021-11-11 at 8 49 43 AM

chawes13 (changes requested):
Can you update the spec as well? You can reference `input-error.test.js` for examples re: extra props and filtering

stories/tables/sortable-table.story.js
```diff
@@ -91,6 +91,13 @@ storiesOf('SortableTable', module)
       <Column name="active" component={CustomCell} />
     </SortableTable>
   ))
+  .add('with custom properties on cell component per column', () => (
+    <SortableTable data={tableData}>
+      <Column name="name" customCellClass="test" />
```
chawes13, line 96: This is a bit misleading since `customCellClass` will be filtered out on the default column component. I think you could add `customCellClass` to the story book example "with custom cell component" instead

test/tables/sortable-table.test.js
```diff
@@ -367,3 +367,12 @@ test('invalid arbitrary props filtered out', () => {
   )
   expect(wrapper.find('table').first().prop('invalidProp')).toBe(undefined)
 })
+
+test('passes extra props to cells', () => {
```
chawes13, line 371: Can you add a test that confirms invalid extra props don't get passed? E.g., `customCellClass`

chawes13 (changes requested):

stories/tables/sortable-table.story.js
```diff
@@ -87,10 +87,17 @@ storiesOf('SortableTable', module)
...
     <SortableTable data={tableData}>
       <Column name="name" />
-      <Column name="age" />
+      <Column name="age"  />
       <Column name="active" component={CustomCell} />
     </SortableTable>
   ))
+  .add('with valid DOM properties on cell component per column', () => (
+    <SortableTable data={tableData}>
+      <Column name="name" tabIndex='tab-index'/>
```
chawes13, line 96: Can you please pass a valid value for tab-index? https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

chawes13:
@aojin Do you have capacity to make these updates?

chawes13 (approved):

Issue #350, opened 2019-06-11
Pass all DOM props down to inner label in InputLabel


dpikt:
As discussed here: https://github.com/LaunchPadLab/lp-components/pull/349#pullrequestreview-248443345

chawes13:
Rationale for _not_ incorporating this here: https://github.com/LaunchPadLab/lp-components/pull/506#issuecomment-1156960937

PR #506, opened 2021-11-17
pass DOM props to inner label tags. also tests

aojin:
[Pass all DOM props down to inner label in InputLabel #350](https://github.com/LaunchPadLab/lp-components/issues/350)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@launchpadlab/lp-components",
-  "version": "5.4.2",
+  "version": "5.4.4",
```
chawes13, line 3: This is actually considered a [breaking change](https://github.com/LaunchPadLab/lp-components/pull/349#issuecomment-501027020), not a bug fix. The issue is included in Milestone v6, so please set your base branch to `v6`

stories/forms/labels/input-label.story.js
```diff
@@ -35,6 +35,12 @@ storiesOf('InputLabel', module)
       requiredIndicator={ '*' }
     />
   ))
+  .add('pass DOM props to inner label tag', () => (
```
chawes13, line 38: Please follow the convention established in this file and start each story description with "with"

stories/forms/labels/input-label.story.js
```diff
@@ -35,6 +35,12 @@ storiesOf('InputLabel', module)
...
+    <InputLabel
+      name="nameOfInput"
+      aria-label="testAriaLabel"
```
chawes13, line 41: Can you swap this out with something like `data-cy`? Passing `aria-label` to a label element is not interpreted correctly by VoiceOver.

chawes13 (changes requested):

chawes13:
In retrospect, I don't think we want this behavior. Since all of these props get passed to the input, this can lead to conflicts (e.g., `data-*` attributes would get added to _both_ the label and the input field)

Issue #460, opened 2020-11-20
DatePicker throws error in StoryBook


chawes13:
`_onBlur` is not a function

![Screen Recording 2020-11-20 at 04 19 57 PM](https://user-images.githubusercontent.com/20426095/99855357-53e5a580-2b4c-11eb-9f50-1f0e83226b1d.gif)

chawes13:
@rgaban Assigning based on the convo we had today!

PR #509, opened 2021-11-19
DateInput in Storybook throws an error

rgaban:
- Issue #[460](https://github.com/LaunchPadLab/lp-components/issues/460): DatePicker throws error in StoryBook

## Author Checklist

- [ ] Add unit test(s)
- [x] Assign dev reviewer


rgaban:
@chawes13 I think I need a walk through on versioning guidelines so I fully understand 😬 another question I had is unit testing. I see that `onBlur` is invoked twice. Is that something I should fix or is it okay to expect that twice?

chawes13 (commented):
Can you update the version please?

This section outlines the general heuristics that we use: https://github.com/LaunchPadLab/opex-public/blob/master/gists/npm-package-guidelines.md#pull-requests-and-deployments

And here's a more formal resource for semantic versioning: https://semver.org/.

Let me know if you have any questions!

chawes13 (approved):
🚀

PR #510, opened 2021-11-19
Add Prettier Code Formatting

danparnella:
Not really resolving an "issue", but formatted code is just so much _prettier_ 😉 .

## Author Checklist

- [x] ~Add unit test(s)~ Nothing to add, but they all still pass
- [x] ~Update documentation (if necessary)~
- [x] ~Add story to storybook (if necessary)~
- [x] Assign dev reviewer


package.json
```diff
@@ -30,7 +30,9 @@
     "test": "jest --coverage",
     "size": "yarn build && size-limit",
     "storybook:build": "build-storybook -o .storybook/static/storybook",
-    "heroku-postbuild": "yarn run storybook:build"
+    "heroku-postbuild": "yarn run storybook:build",
+    "format": "prettier --write \"src/**/*.+(js|jsx|json|scss)\"",
+    "format-storybook": "prettier --write \".storybook/**/*.+(js|jsx|json|scss)\""
```
danparnella, line 35: I know we don't normally include styles, but is there a reason for that?

package.json
```diff
@@ -95,6 +98,7 @@
     "react": "^16.8.0"
   },
   "husky": {
-    "pre-commit": "yarn run docs && git add docs.md"
-  }
+    "pre-commit": "yarn run docs && yarn run format && yarn run format-storybook && git add"
```
danparnella, line 101: Added formatting to pre-commit. If this is a good idea, should we also add linting?

src/routes/index.js
```diff
@@ -1,2 +1,4 @@
+/* eslint-disable import/no-named-as-default */
 export AuthorizedRoute from './authorized-route'
 export UnauthorizedRoute from './unauthorized-route'
+/* eslint-disable import/no-named-as-default */
```
danparnella, line 4: Do you know of a better way to handle this? The issue is they're wrapped in an HOC when being exported so I think these names should technically be different. We could change the names, keep this single instance of disabling the rule, or we could ignore this rule globally.

src/tables/components/table-row.js
```diff
@@ -47,9 +47,8 @@ function TableRow({
         const onColClick = column.disabled ? noop : () => onClick(rowData)
         return (
           <CellComponent
+            key={key}
             {...{
-              // eslint-disable-line
-              key,
```
danparnella: Seemed like a small enough change to avoid needing to disable linting 😄 .

src/routes/index.js
chawes13: I believe this was fixed in `v6` https://github.com/LaunchPadLab/lp-components/blob/7da7efea655a0716e5988eb676be68472f1d5106/src/routes/index.js

danparnella: Ah, what's the process for that. Should I be branching off of that instead, or wait until that version is released?

package.json
chawes13: This was a remnant of not wanting to interfere with how the design team does their work. I think we can update it

src/routes/index.js
chawes13: I would prefer to run formatting on v6 right before it's merged. I don't want to have to deal with the conflicts that will result from code shifting around in v5.

chawes13 (changes requested):
Could you port over the config changes and run this branched off of the `v6` milestone branch? We have implemented several breaking changes already and I'd prefer to limit our merge conflict headaches!

chawes13:
@danparnella I'm going to port over these changes in **v7.0.1**, so that it's easier for devs to analyze the code changes from `v6@latest` vs. `v7.0.0` without being distracted by the formatting changes

Issue #463, opened 2021-01-06
Allow for caption to be specified for Table components


chawes13:
We should be able to specify the text _or_ component to use as a [`<caption>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/caption) element in our Table components. We could leverage a similar pattern as established for label / labelComponent in [LabeledField](https://github.com/LaunchPadLab/lp-components/blob/0ea9404056f8bece2fd3fe84e09de57222a7c985/src/forms/labels/labeled-field.js#L96), i.e., accept a `caption` string prop and a `captionComponent` prop that defaults to `({ caption }) => <caption>{caption}</caption>`.

This is related to research by @mwislek in addressing #324

dpikt:
👍good callout. I think we could just make this a node prop that will be inserted into the tag since it wouldn't need access to any "internal" props/state (I think):

```
caption="My Table"

...

caption={<p>My<strong>important</strong>table</p>}
```

chawes13:
Is there a pattern for accepting `Node|String`? Passing in text seems like the primary use case, and `caption={<>My Table</>}` feels a little clunky

dpikt:
A string is a node, so it would work that way by default 🙂 My example above was an "inclusive or" of the ways it could be used.

dpikt:
To be clear, I mean node as in `PropTypes.node`, not a DOM node.

chawes13:
Awesome, thanks for the clarification!

PR #513, opened 2021-11-19
add caption component for table

jhp0621:
Allow table to have a caption specified: https://github.com/LaunchPadLab/lp-components/issues/463

## Author Checklist

- [ ] Add unit test(s)
- [x] Assign dev reviewer

### DIFF
--- a/src/tables/components/table-caption.js
+++ b/src/tables/components/table-caption.js
@@ -1,0 +1,7 @@
+import React from "react";
+
+function CaptionComponent({ caption, className }) {
+  return <caption className={className}>{caption}</caption>;
+}
+
+export default CaptionComponent;
--- a/src/tables/sortable-table.js
+++ b/src/tables/sortable-table.js
@@ -1,9 +1,13 @@
-import React, { useMemo, useState } from 'react'
-import PropTypes from 'prop-types'
-import { getColumnData, Types } from './helpers'
-import { TableHeader as DefaultHeader, TableRow as Row } from './components'
-import { filterInvalidDOMProps, get, noop, orderBy } from '../utils'
-import classnames from 'classnames'
+import React, { useMemo, useState } from "react";
+import PropTypes from "prop-types";
+import { getColumnData, Types } from "./helpers";
+import {
+  TableHeader as DefaultHeader,
+  TableRow as Row,
+  TableCaption as Caption,
+} from "./components";
+import { filterInvalidDOMProps, get, noop, orderBy } from "../utils";
+import classnames from "classnames";
 
 /**
  * A component for displaying sortable data in a table.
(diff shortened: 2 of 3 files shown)


src/tables/components/table-caption.js
```diff
@@ -0,0 +1,7 @@
+import React from "react";
+
+function CaptionComponent({ caption, className }) {
+  return <caption className={className}>{caption}</caption>;
+}
+
+export default CaptionComponent;
```
chawes13, line 7: By convention we match the name of the file with the name of the component. E.g., https://github.com/LaunchPadLab/lp-components/blob/master/src/tables/components/table-header.js

src/tables/components/table-caption.js
```diff
@@ -0,0 +1,7 @@
+import React from "react";
+
+function CaptionComponent({ caption, className }) {
+  return <caption className={className}>{caption}</caption>;
```
chawes13, line 4: We don't have prettier on this repo (yet!!! #510), so could you remove the semicolons manually please?

src/tables/sortable-table.js
```diff
@@ -47,39 +51,43 @@ const propTypes = {
...
-  initialSortPath: '',
+  initialSortPath: "",
   initialSortFunc: null,
   initialValueGetter: null,
-}
+};
 
 function getInitialSortControls(initialColumn, columns) {
-  if (!initialColumn) return defaultControls
+  if (!initialColumn) return defaultControls;
```
chawes13, line 75: It would be great if you could undo the formatting applied. I think your global prettier settings must be different than LaunchPad's

src/tables/sortable-table.js
```diff
@@ -47,39 +51,43 @@ const propTypes = {
   onChange: PropTypes.func,
   rowComponent: Types.component,
   headerComponent: Types.component,
-}
+  ...Caption.propTypes,
```
chawes13, line 54: I don't think this is doing anything? You aren't defining `TableCaption.propTypes` in the component, so I think this will always spread nothing.

Additionally, if you had defined `propTypes` as
```js
const propTypes = {
  className: PropTypes.string,
  caption: PropTypes.string,
}
```
then this would be overridden by the inclusion of `caption: PropTypes.string` on line 55 (and override `className: PropTypes.string` on line 44).

src/tables/sortable-table.js
```diff
@@ -92,101 +100,123 @@ function SortableTable({
...
-      sortFunc: newSortFunc
-    })
-  }
+    if (onChange)
+      onChange({
+        ascending: newAscending,
+        sortPath: newSortPath,
+        sortFunc: newSortFunc,
+      });
+  };
```
chawes13, line 170: Did anything change here? It's hard to tell 🙈

src/tables/sortable-table.js
```diff
@@ -47,39 +51,43 @@ const propTypes = {
   onChange: PropTypes.func,
   rowComponent: Types.component,
   headerComponent: Types.component,
-}
+  ...Caption.propTypes,
+  caption: PropTypes.string,
```
chawes13, line 55: Per [David's request](https://github.com/LaunchPadLab/lp-components/issues/463#issuecomment-756344487), we should make this a `node`

src/tables/sortable-table.js
```diff
@@ -92,101 +100,123 @@ function SortableTable({
...
+  };
 
   return (
-    <table className={classnames(className, { 'sortable-table': !disableSort })} {...filterInvalidDOMProps(rest)}>
-      <thead>
-        <tr>
-          {columns.map((column, key) => {
-              const Header = column.headerComponent || headerComponent || DefaultHeader
+    <>
+      <CaptionComponent caption={caption} />
```
chawes13, line 174: As a `node`, you could just always render the caption. E.g.,
```jsx
<caption>{caption}</caption>
```
However, since we'd be introducing a new change this could impact styling on existing projects. I believe this is tagged to the `v6` milestone, so please modify the base branch to `v6` and add a note in the [migration guide](https://github.com/LaunchPadLab/lp-components/blob/7da7efea655a0716e5988eb676be68472f1d5106/migration-guides/v6.0.0.md)

chawes13 (changes requested):

jhp0621:
closing in favor of https://github.com/LaunchPadLab/lp-components/pull/540

Issue #436, opened 2020-07-17
aria-labelled by for LabeledField


inveterateliterate:
from pa11y-ci:

 This fieldset element does not have a name available to an accessibility API. Valid names are: legend element, aria-label undefined, aria-labelledby undefined.

dpikt:
@chawes13 was the second point what you fixed in https://github.com/LaunchPadLab/lp-components/pull/384? (unreleased)

inveterateliterate:
and if so @dpikt  what's the release plan for v4? we can ignore this in our CI until then

dpikt:
Well, it's been open since Jan 2019 😵I can prioritize getting it merged next week. I was trying to finish https://github.com/LaunchPadLab/lp-components/issues/410 first but I don't think that's happening any time soon.

inveterateliterate:
it's not urgent! I know you're swamped. I will put a flag to ignore in CI for now

dpikt:
https://github.com/LaunchPadLab/lp-components/pull/311

chawes13:
@dpikt This issue would only be partially addressed by #384 (for radio and checkbox groups). While it does appear that a fieldset with a single input is valid HTML, it's not recommended according to WebAIM (excerpt below).

> Fieldset and legend should only be used when a higher-level label is necessary. Single checkboxes or basic radio buttons that make sense from their labels alone do not require fieldset and legend.

Another source ([UK Gov A11y blog](https://accessibility.blog.gov.uk/2016/07/22/using-the-fieldset-and-legend-elements/)), states:

> You should not use the `<fieldset>` and `<legend>` when: You have a single form field that asks for a single piece of information.

I believe that this indicates that `<fieldset>` is not the appropriate element to be wrapping inputs **except for** Radio and Checkbox groups. Since `v4` is unreleased, it's possible that we could include a change that swaps out the element for a `div`. I wonder how other design systems handle this - it might be worth doing more design research to figure out the most appropriate element.

inveterateliterate:
thanks @chawes13  and @dpikt  for now I just made a version of `LabeledField` in **prosci-components** and changed `fieldset` to `div` (won't link here because the repo is limited). I will revert back once v4 is published!

jhp0621:
[...]

PR #514, opened 2021-11-19
Input Label Accessibility Updates 

danparnella:
- Resolves https://github.com/LaunchPadLab/lp-components/issues/436
- ~~Plan is also to add aria props to any `input` based component~~
- Would also like to make sure the fieldset/legend setup is correct for checkbox and radio groups

## Author Checklist

- [x] Add unit test(s)
- [x] Assign dev reviewer


src/forms/inputs/input.js
```diff
@@ -25,6 +25,8 @@ import {
  * @param {Object} input - A `redux-forms` [input](http://redux-form.com/6.5.0/docs/api/Field.md/#input-props) object
  * @param {Object} meta - A `redux-forms` [meta](http://redux-form.com/6.5.0/docs/api/Field.md/#meta-props) object
  * @param {String} [type] - A string to specify the type of input element (defaults to `text`)
+ * @param {String} [ariaLabel] - A string to specify the label of the input element to assistive technologies when a standard label is not provided (will be overriden by `ariaLabelledby`)
+ * @param {String} [ariaLabelledby] - A string that corresponds to the `id` of one or more elements used as the label of the input element for assistive technologies when a standard label is not provided (will override `ariaLabel`)
```
danparnella, line 29: @chawes13 Before I go too crazy with this (adding these to all the other `input` based components), does this look like the right pattern to follow? I don't see too many aria type props spelled out this way, but I wonder if we should be doing that.

chawes13: @danparnella I'm not sure I follow the need for this change. I can pass it straight through the `Field` component and it gets added correctly.

E.g., https://codesandbox.io/s/tender-marco-giusn?file=/src/Form.js

chawes13:
@danparnella Can you do an analysis on how this will impact styling (both in storybook and in the client template)?

src/forms/inputs/input.js
danparnella: 🤔 Hopefully, I was just following what we had done in artlook: https://github.com/parliamentdata/artlook-client/pull/808/files#diff-431ff99478b73c952842cc3b24c9fe645eda70de5f5a98eb19cb472a64c3a694 and that we can just pass it through like you said. I'll go down that path and revert these changes!

.storybook/styles/application.scss
```diff
@@ -48,10 +48,8 @@ $output-bourbon-deprecation-warnings: false;
 @import 'components/progress-circle';
 @import 'components/login';
 @import 'components/pagination';
-@import 'components/sidebar-content';
```
danparnella, line 51: I noticed a couple style files were no longer being used. Not sure if they should still be there for reference?

.storybook/styles/components/_forms.scss
```diff
@@ -1,7 +1,8 @@
 /*-----------------------
-Fieldset
+Field Wrappers
 -----------------------*/
-fieldset {
+fieldset,
+.field-wrapper {
```
danparnella, line 5: My hope is that this might ease some of the style changes (just swapping `fieldset` for `.field-wrapper` in stylesheets). Hopefully most projects can just find and replace in their stylesheets without too much auditing.

.storybook/styles/components/_forms.scss
```diff
@@ -85,6 +86,19 @@ Input Icon
...
+
+    i {
+      @include position(absolute, 3px null null 10px);
+      background-repeat: no-repeat;
+      background-size: 15px;
+      display: inline-block;
+      height: 20px;
+      width: 20px;
+    }
+  }
```
danparnella, line 101: This was included further down in a duplicate "Input Icon" section.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@launchpadlab/lp-components",
-  "version": "8.0.0",
+  "version": "9.0.0",
```
danparnella, line 3: I figured I'd hold off on writing a migration guide until the code changes are approved.

src/forms/inputs/checkbox-group.js
```diff
@@ -96,11 +95,6 @@ const defaultProps = {
   options: [],
 }
 
-function CheckboxGroupLegend({ name, label }) {
-  if (label === false) return null
-  return <legend>{label || convertNameToLabel(name)}</legend>
-}
```
danparnella, line 102: This is now basically the default for `LabeledFieldset`.

src/forms/inputs/dropdown-checkbox-group.js
```diff
@@ -59,13 +59,14 @@ function DropdownCheckboxGroup(props) {
...
-    <fieldset>
-      <InputLabel {...{ label, name }} />
+    <div>
+      <FieldsetLegend {...{ label, name }} id={inputId} />
       <DropdownSelect selectedValues={value} className="checkboxes">
-        <CheckboxGroup {...props} label={false} />
+        <CheckboxGroup {...props} label={false} ariaLabelledby={inputId} />
       </DropdownSelect>
-    </fieldset>
+    </div>
```
danparnella, line 69: The checkbox group is actually the `fieldset` so `div` should be correct here and since a label/legend isn't provided, I'm using the `id` on the "outer" legend to provide the aria labelledby.

src/forms/labels/fieldset-legend.js
```diff
@@ -0,0 +1,141 @@
...
+import { useToggle } from '../../utils'
+
+/**
+ *
+ * A dynamic legend associated with a group of inputs within a fieldset.
+ *
+ * This component is used within {@link LabeledFieldset}, and therefore is incorporated into most fieldset grouped input components ({@link RadioGroup}, {@link CheckboxGroup}) by default.
+ *
+ * The text of the legend is set using the following rules:
+ * - If the `label` prop is set to `false`, the legend is hidden completely
```
danparnella, line 14: I kept debating whether to change "label" fully to "legend", but I was thinking for backwards compatibility (potentially in code, but also in terms of consistency mentally), sticking with `label` as the prop might be better. It's also still a "label", just a different kind of one. BUT, that's just my two cents.

danparnella:
@jhp0621 Tagging you for visibility!

src/forms/inputs/radio-group.js
```diff
@@ -133,16 +127,12 @@ function RadioGroup(props) {
   } = omitLabelProps(props)
   const optionObjects = serializeOptions(options)
   return (
-    <LabeledField
-      className={className}
-      labelComponent={RadioGroupLegend}
-      {...props}
-    >
-      {optionObjects.map((option, i) => {
+    <LabeledFieldset className={className} {...props}>
```
jhp0621, line 130: @danparnella what happens if `label` is `false`? Pretty certain `<fieldset>` must have a unique `<legend>`; if a radio group has `label` set to `false`, I believe we would still be rendering a `<fieldset>` but without a `<legend>`?

danparnella: Yep, that's correct. If the design requires `label={false}`, then the developer should use `aria-labelledby` to specify which element to use as the "legend". It might be good for me to add that as an example in the story though.

chawes13: It sounds like Ji ran into some issues where `aria-labelledby` on a fieldset still fails accessibility checks because a legend is **required** when using a `fieldset` ([reference](https://www.w3.org/TR/WCAG20-TECHS/H71.html)). An alternative might be to have `<div role="group" aria-labelledby="">` in this instance.

The reference above is a _technique_ which isn't _required_ to meet WCAG, but is rather a guideline. We might need to dig more into the actual Success Criterion to see if this is a strict requirement or rather a strict interpretation on the accessibility tool. Maybe we could do testing with SiteImprove and Axe's accessibility testing extensions?

@jhp0621 Am I remembering correctly? Do you have more information/resources that you can share?

chawes13: I just did some testing [here](https://q6ownu.csb.app/), and the output is the exact same on VoiceOver. @jhp0621 maybe something to dig into a little deeper with the EA report. Does it list which success criterion are failing?

.storybook/styles/application.scss
chawes13: I'm fine with deleting them!

.storybook/styles/components/_forms.scss
chawes13: Can you add this documentation to a migration guide?

src/forms/inputs/dropdown-checkbox-group.js
chawes13: If we aren't using a `fieldset`, is `legend` the semantically appropriate tag? Or would `p`/`span` suffice?

Looking at the specs, the only Permitted parent is a `fieldset` https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend#technical_summary

stories/forms/labels/labeled-fieldset.story.js
```diff
@@ -0,0 +1,113 @@
...
+        hideErrorLabel: true,
+      }}
+    >
+      <Checkbox input={inputProps} meta={{}} />
+    </LabeledFieldset>
+  ))
+  .add('with custom label', () => {
+    // eslint-disable-next-line
+    const CustomLabel = ({ onClickLabel }) => (
+      <legend onClick={onClickLabel} htmlFor="inputName">
```
chawes13, line 70: Can we pass a different custom prop (or use it in a different way)? I don't want to encourage inappropriate `onClick` handler usage

stories/forms/labels/labeled-fieldset.story.js
```diff
@@ -0,0 +1,113 @@
...
+          onClickLabel: action('Custom Label Clicked'),
+        }}
+      >
+        <Checkbox input={inputProps} meta={{}} />
+      </LabeledFieldset>
+    )
+  })
+  .add('with custom error', () => {
+    const CustomError = (props) => (
+      <span id="inputError">
```
chawes13, line 91: Can we change this to `inputNameError` or use the generate input error id util here? I don't want to encourage the wrong usage of a custom error component (the `aria-describedby` that's added to the input won't match this `id`)

test/forms/labels/fieldset-legend.test.js
```diff
@@ -0,0 +1,82 @@
...
+test('when label provided - renders a legend with content equal to string', () => {
+  const wrapper = mount(<FieldsetLegend name={name} label="foo" />)
+  expect(wrapper.find('legend').text()).toEqual('foo')
+})
+
+test('when children are provided, renders a legend with content equal to children', () => {
+  const onClick = jest.fn()
+  const wrapper = mount(
+    <FieldsetLegend name={name}>
+      Are you <span onClick={onClick}>sure</span>?
```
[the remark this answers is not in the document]
chawes13: Or remove it? Doesn't seem relevant to the test 🤔

test/forms/labels/labeled-fieldset.test.js
```diff
@@ -0,0 +1,185 @@
...
+test('wraps children in fieldset', () => {
+  const Wrapped = () => <input name="test" />
+  const props = { input: { name: 'foo' }, meta: {} }
+  const wrapper = mount(
+    <LabeledFieldset {...props}>
+      <Wrapped />
+    </LabeledFieldset>
+  )
+  expect(wrapper.find('fieldset').exists()).toEqual(true)
+  expect(wrapper.find('fieldset').hasClass('error')).toEqual(false)
```
chawes13, line 14: This doesn't align with the test description 🤔 . Can we create a separate test for this?

danparnella:
Closing in favor of #595

Issue #518, opened 2021-12-01
Don't overwrite modal-inner when classname is passed


chawes13:
Since there are default styles included related to this styling, we shouldn't automatically override it.

Workaround: Include `modal-inner` in the passed in className. E.g.,
```jsx
<Modal className="modal-inner support-modal" />
```

Amihozar:
Hi @chawes13 , May I work on this?

PR #519, opened 2021-12-03
Resolves Issue518 - Don't overwrite modal-inner when classname is passed

Amihozar:
Resolving #518 by customizing Modal className property inspired by [React Modal buildCalssName](https://github.com/reactjs/react-modal/blob/702ff620924af916cf08223871d563efd5a96294/src/components/ModalPortal.js#L331)

## Author Checklist

- [x] Add unit test(s)
- [ ] Assign dev reviewer

### DIFF
--- a/src/modal.js
+++ b/src/modal.js
@@ -39,6 +39,7 @@ import { isServer } from './utils'
  */
 
 const propTypes = {
+  className: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
   onClose: PropTypes.func.isRequired,
   hideCloseButton: PropTypes.bool,
   children: PropTypes.node,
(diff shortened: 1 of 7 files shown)


src/modal.js
```diff
@@ -55,15 +56,28 @@ function getRootElement() {
...
+  // No className specefied or className property is empty
+  if (!additional)
+    return main
+  // A nice and clean className, nothing special to do
+  if (typeof additional === 'string')
+    return `${main} ${additional}`
+  // react-modal will handle it
+  // see http://reactcommunity.org/react-modal/styles/classes/
+  if (typeof additional === 'object')
+    return additional
```
chawes13, line 69: Would this override `main`?

chawes13:
@Amihozar Thanks for the PR! If you use the word "resolve", "resolves", or "resolved", then the issue will be linked and automatically close when this PR is merged. However, "resolving" is not one of those key words. More info here: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

src/modal.js
Amihozar: > 

Yes it does.
Passing an object to className will override both main and ReactModal__Content classes.

chawes13 (changes requested):
Looking great! Can you update your base branch to `LaunchPadLab:v6` and update the migration guide accordingly? There might be apps that rely on the behavior to overwrite the class, so I don't think we should include it in `v5` (i.e., the current major version)

chawes13:
This PR is deprecated by https://github.com/LaunchPadLab/lp-components/pull/554

PR #525, opened 2022-01-18
Modal close refactor

### LINKED ISSUE #517: Allow Modal's onClose to be null and derive other props from it
Instead of having to pass in a function and several related props, derive the props from the lack of having a close function supplied. It's common for `hideCloseButton` to be the only prop that is considered, which can lead to defects with the user still being able to close a modal via the keyboard or pressing the overlay.

```jsx
// Current State
<Modal
  onClose={() => setModalShown(false)}
  hideCloseButton={!canClose}
  shouldCloseOnEsc={canClose}
  shouldCloseOnOverlayClick={canClose}
/>

// Proposed
<Modal onClose={canClose ? () => setModalShown(false) : null} />
```

```jsx
// Proposed implementation
function Modal({ onClose, children, ...rest }) {
    return (
    <ReactModal
      isOpen
      onRequestClose={onClose ?? noop}
      portalClassName="modal"
      className="modal-inner"
      overlayClassName="modal-fade-screen"
      bodyOpenClassName="modal-open"
      appElement={getRootElement()}
      ariaHideApp={isServer()} // Opt out of setting appElement on the server.
      shouldCloseOnEsc={!!onClose}
      shouldCloseOnOverlayClick={!!onClose}
      {...rest}
    >
      <div className="modal-content">{children}</div>
      {!!onClose  && (
        <>
          <button
            onClick={onClose}
            className="modal-close"
            aria-label="Close Modal"
          >
            ×
          </button>
        </>
      )}
    </ReactModal>
  )
}
```

chawes13:
Resolves #517 

## Author Checklist

- [X] Add unit test(s)
- [X] Update version in package.json (see the [versioning guidelines](https://github.com/LaunchPadLab/opex-public/blob/main/gists/npm-package-guidelines.md#pull-requests-and-deployments))
- [X] Add story to storybook (if necessary)
- [x] Assign dev reviewer


src/modal.js
```diff
@@ -57,23 +57,27 @@ function getRootElement() {
 
 // A wrapper for react-modal that adds some styles and a close button.
 // See https://github.com/reactjs/react-modal for usage.
-function Modal({ onClose, hideCloseButton, children, ...rest }) {
+function Modal({ onClose, preventClose, children, ...rest }) {
+  const canClose = !preventClose
   return (
     <ReactModal
-      isOpen
+      isOpen={true}
```
bhennes2, line 64: I follow the updated `preventClose` prop but not sure I follow why `isOpen` is always `true`? Seems like it can be overridden by the `{...rest}`.

bhennes2: Would it be more clear to set `defaultProps` value for `isOpen` and then this line would be `isOpen={isOpen}`?

bhennes2 (commented):
Looking good but added a few comments for clarification.

src/modal.js
chawes13: @bhennes2 I'm not actually changing any functionality here. `isOpen` is syntactic sugar for `isOpen={true}`, since props default to true. However, the React documents actually recommend _against_ this (hence the change).

https://reactjs.org/docs/jsx-in-depth.html#props-default-to-true

bhennes2: 👍 👍

bhennes2 (approved):

src/modal.js
chawes13: I like your suggestion! I'll make that change in addition to the changes made for `preventClose` in this PR

PR #540, opened 2022-03-19
render caption in table if passed in as a prop

jhp0621:
Allow table to have a caption specified: https://github.com/LaunchPadLab/lp-components/issues/463

## Author Checklist

- [ ] Add unit test(s)
- [ ] Assign dev reviewer


jhp0621:
opening a new PR since it was gonna be hard to undo formatting in https://github.com/LaunchPadLab/lp-components/pull/513

src/tables/sortable-table.js
```diff
@@ -148,41 +151,49 @@ function SortableTable({
   }
 
   return (
-    <table className={classnames(className, { 'sortable-table': !disableSort })} {...filterInvalidDOMProps(rest)}>
-      <thead>
-        <tr>
-          {columns.map((column, key) => {
-              const Header = column.headerComponent || headerComponent || DefaultHeader
+    <>
+      {caption && <caption>{caption}</caption>}
```
jhp0621, line 155: this is the only change in return

migration-guides/v6.0.0.md
```diff
@@ -269,4 +271,25 @@ Replace `hideCloseButton` with `preventClose`. If you still need the modal to cl
...
+
+## 13. `<Table />` and `<SortableTable />` components now render a `<caption>` element with a `caption` prop it receives
+
+`caption` is a node prop and defaults to null. <caption> html tag is only rendered if `caption` is present.
+```jsx
+// Before
+<caption>My Awesome Table</caption>
+<Table>
+  Table content
+</Table>
```
chawes13, line 284: I don't think this is an appropriate Before example. The caption tag must be rendered **within** the table tag
Image 2022-06-14 at 9 46 37 AM

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/caption

src/tables/sortable-table.js
chawes13: This needs to be rendered as the first child of the `<table>`. See previous comment for source

chawes13 (changes requested):
Can you also add tests?

chawes13:
@jhp0621 I decided against including information in the migration guide because this is not a breaking change. I also updated the placement and added a few tests. Let me know if you have any questions!

chawes13 (approved):

PR #543, opened 2022-04-07
Modernize lp-components

mwislek:
This PR adds Node 16 support to lp-components. Additionally many of this module's dependencies were updated to the appropriate compatible version. Note the following updates were **not** made and should be addressed by future PRs:

- While the `storybook` dependencies were updated, the files used to configure `storybook` will soon be deprecated. That should be addressed along with a upgrade to a `storybook` version that supports `web pack` ^5.0.0. 
- `react-router` was not upgraded. The support for before/change hooks used in `AuthorizedRoute` and `UnauthorizedRoute` components is obsolete and should be refactored, like using React hooks (`useEffect()`)
- `DateInput` was not upgraded. This component uses `moment` under the hood. Newer versions of `react-datepicker` eliminate `moment` for `date-fns`. The `selected` prop should be updated to accept a JS `Date` rather than a `moment()`.
- The Babel upgrade causes a lot of warnings to be generated when `yarn storybook` is run. Several different attempts were made to resolve these warnings. These warnings should be corrected if/when the underlying `babel-preset` is upgraded and the `storybook` configuration mechanisms are updated.
- An attempt was made to upgrade `jest` but that failed. The current use of `mount()` seems to require a JSDOM installation.

## Author Checklist

- [ ] Add unit test(s). No tests were added for this PR. Instead, the existing test suite was executed to verify that no errors were introduced.
- [x] Update version in package.json (see the [versioning guidelines](https://github.com/LaunchPadLab/opex-public/blob/main/gists/npm-package-guidelines.md#pull-requests-and-deployments)). Given that support for earlier Node versions was eliminated, this version is considered a MAJOR release.
- [x] Update documentation (if necessary), Added migration information.
- [ ] Add story to storybook (if necessary) Note: not necessary.
- [x] Assign dev reviewer


.storybook/addons.js
```diff
@@ -2,4 +2,4 @@ import '@storybook/addon-actions/register'
 import '@storybook/addon-a11y/register'
 import '@storybook/addon-storysource/register'
 import '@storybook/addon-viewport/register'
-import '@storybook/addon-knobs/register'
+import '@storybook/addon-knobs/dist/register'
```
mwislek, line 5: This plug-in was actually deprecated though support is still provided. The upgrade, for some reason, moved the `register` function to the `dist` folder.

.storybook/styles/application.scss
```diff
@@ -15,8 +15,8 @@
  */
 
 // Imports
-@import "~bourbon/app/assets/stylesheets/bourbon";
-@import "~bourbon-neat/app/assets/stylesheets/neat";
+@import "../../node_modules/bourbon/core/bourbon";
```
mwislek, line 18: All of these changes in the `.scss` files arose from a later version of `node-sass`. All of these changes are comparable to similar changes made to the stylesheets added to `the client template.

.storybook/webpack.config.js
```diff
@@ -24,9 +24,9 @@ module.exports = async ({ config }) => {
         options: {
           sassOptions: {
             includePaths: [
-              datePickerIncludePaths,
-              bourbonIncludePaths,
-              neatIncludePaths,
+              ...datePickerIncludePaths,
```
mwislek, line 27: Had to make this change to the client template too. Arrays of arrays are not supported.

migration-guides/v6.0.0.md
```diff
@@ -0,0 +1,4 @@
+# v6.0.0 Migration Guide
+
+This version introduces support for [Node](https://nodejs.org/en/) version of ^16.0.0. No `lp-components` interface changes were introduced in this version, though many of the underlying dependencies used to build the `lp-components` have been updated. This version also eliminates support for earlier Node versions. [Client template](https://github.com/LaunchPadLab/client-template) version ^6.0.0 is compatible with this `lp-components` version.
+
```
mwislek, line 4: Let me know if you think there's more to say here.

package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "@launchpadlab/lp-components",
-  "version": "5.4.8",
+  "version": "6.0.0",
   "engines": {
-    "node": "^8.0.0 || ^10.13.0 || ^12.0.0"
+    "node": "^16.0.0"
```
mwislek, line 5: I personally saw no reason to support the earlier Node versions and, so, removed them. Let me know if you disagree. I think this may still work with Node 12 but I saw no reason to support earlier Node version.

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
mwislek, line 1: Needed for husky upgrade...

.nvmrc
```diff
@@ -1 +1 @@
-v12.18.2
+v16.14.2
```
mwislek, line 1: 🚀

.size-limit.js
```diff
@@ -1,7 +1,7 @@
 module.exports = [
   {
     path: 'lib/index.js',
-    limit: '200 KB',
+    limit: '250 KB',
```
mwislek, line 4: Had to increase the limit as the upgrade added some additional overhead. We're at about 238K now.

test/setup.js
```diff
@@ -1,3 +1,7 @@
+// Required since lp-hoc > lp-requests uses async/await
+import 'core-js/stable'
```
mwislek, line 2: Needed for Babel update, similar to change made in the client template.

.storybook/addons.js
chawes13: Do you think we should drop knobs support altogether? Or should we consider that out of scope for this PR?

package.json
chawes13: Node 14 still has a year left of maintenance support. Ideally we'd support both to allow for more incremental upgrades. I wouldn't invest more than an hour trying to troubleshoot issues that come up though

chawes13, line 5: Should we make this `^16.14.2`?

package.json
```diff
@@ -40,61 +40,57 @@
...
+    "react-dom": "16.14.0",
     "recompose": "^0.30.0",
-    "redux": "^4.0.0",
-    "sass-loader": "^8.0.2",
-    "style-loader": "^1.0.0",
+    "redux": "^4.1.2",
+    "regenerator-runtime": "^0.13.9",
+    "sass-loader": "10.2.1",
+    "size-limit": "^7.0.8",
+    "style-loader": "2.0.0",
```
chawes13, line 90: Is there a specific reason why this went from flexible to pinned?

package.json
```diff
@@ -40,61 +40,57 @@
...
+    "eslint": "^8.12.0",
+    "husky": "^7.0.4",
     "jest": "^25.1.0",
     "jest-canvas-mock": "^2.1.1",
-    "node-sass": "^4.11.0",
-    "react": "^16.8.0",
-    "react-dom": "^16.2.0",
+    "node-sass": "6.0.1",
+    "react": "16.14.0",
+    "react-dom": "16.14.0",
```
chawes13, line 84: Is there a specific reason why these went from flexible to pinned?

package.json
```diff
@@ -40,61 +40,57 @@
...
     "recompose": "^0.30.0",
-    "redux": "^4.0.0",
-    "sass-loader": "^8.0.2",
-    "style-loader": "^1.0.0",
+    "redux": "^4.1.2",
+    "regenerator-runtime": "^0.13.9",
+    "sass-loader": "10.2.1",
```
chawes13, line 88: Is there a specific reason why this went from flexible to pinned?

.storybook/addons.js
mwislek: Hmm. Good question. I couldn't find any explicit configuration of knobs when I looked so I'll look into removing it, since it looks like it was added "for the future".

package.json
mwislek: 😄 When one runs `yarn upgrade X@latest` the latest version is written to the `package.json`. All of these were upgraded with `yarn upgrade` so that the `yarn.lock` file remained correct.  I must admit I was confused on the "philosophy" when I started working on these templates. I originally changed all of these to minor versions to allow flexibility and then reverted to what was originally present, which seemed to be a mix of pinned and flexible versions. Let me know what you think. I could see updating these to at least a pinned minor version (middle number). So, for example, `react-dom` would remain as is but `documentation` would be updated to `^13.2.0` or even `^13.2`.

package.json
mwislek: Related to the conversation below, I could see at least pinning to a minor version (middle number). So, `^16.14` or `^16.14.0`.

mwislek: Done. I updated to allow `^14.19` (latest Node 14 version) and `^16.14` (latest Node 16 version)

package.json
chawes13: Can you walk me through the rationale for pinning to a minor version?

package.json
```diff
@@ -40,61 +40,56 @@
...
+    "bourbon": "^7.2.0",
+    "bourbon-neat": "^4.0.0",
+    "core-js": "^3.21.1",
+    "css-loader": "5.2.7",
+    "documentation": "^13.2.5",
     "enzyme": "^3.2.0",
     "enzyme-adapter-react-16": "^1.1.0",
-    "eslint": "^6.2.1",
-    "husky": "^4.2.1",
+    "eslint": "^8.12.0",
```
chawes13, line 77: I have a vague memory of eslint v8 not working when you tried to update the ionic template (https://github.com/LaunchPadLab/ionic-client-template/pull/19)?

That might be what's causing this build to fail. Could you please investigate? 
Image 2022-04-18 at 9 52 35 PM

migration-guides/v6.0.0.md
chawes13: I think this needs to get tweaked slightly now to accommodate the ^14 allowance

mwislek: Good catch! Updated

package.json
mwislek: "Pinning to a minor version" may have been a misstatement on my part. I meant pinning the minor version element but leaving the patch version as 0. That would mean any version within that minor version would be allowed. Again, I'm not sure I understand the philosophy. I assume we're worried that major versions may introduce incompatibilities and so don't allow that to change. Minor versions should be less risky but, I feel, still should be explicitly controlled by us. The patch version should be able to be updated without issue. So that would mean "flexible" versions for the patch version only. Honestly, not sure how to achieve that using `yarn` without explicitly editing the `package.json` file after doing an upgrade to unpin the patch version. Maybe we need to discuss.

mwislek: I now see what you're referring to now. The carat (^) is missing from a few of these. I'll fix all of those but leave the version numbers as is for now, unless you feel otherwise.

chawes13: If you want to pin to the minor version, you can use `~` (refer to https://stackoverflow.com/a/22345808). However, I would prefer to stick with being more flexible by pinning the major version using the caret. If a library follows [semver](https://semver.org/) appropriately, I would think that we would welcome minor updates.

All that being said, it doesn't really matter for this library because it has a `yarn.lock` file! Where this really comes into play is the client-template, which deletes the `yarn.lock` and installs whatever is the latest in the acceptable range defined in `package.json`. Again, happy to discuss if there are concerns about being flexible with minor version updates

mwislek: Thanks. Yeah, agree that is less of an issue here but I didn't intend to pin to a version either. That happened when I upgraded certain libraries to their "non-latest" values. Glad you noticed that! Anyway, fixed.

package.json
mwislek: God, I can't believe you remembered this! I did not. Anyway, you're (I was??) right: our ESLint settings didn't work with the latest version of `eslint`. Downgraded to 7.32.0, which seems to work when `yarn lint` is invoked.

chawes13 (approved):
🚀

PR #544, opened 2022-04-19
Remove react router

### LINKED ISSUE #416: Remove custom Route components
We don't use them in our latest apps, which use React Router >v4.

However, they are used in some legacy projects using RR v3, so until those are upgraded or no longer supported this may be an unfeasible change.

chawes13:
Resolves #416 

This has been on my todo list for a long time and felt like an appropriate inclusion with the modernization effort (especially given how old react-router@3 is).

## Author Checklist

- [ ] Add unit test(s)
- [X] Assign dev reviewer


mwislek (approved):
Looks good. I assume that the documentation generator updated all of those "table of contents" links, not you. Otherwise, we might look at changing the way that works.

PR #573, opened 2022-07-06
Hotfix: Fix storybook build

chawes13:
## Items Addressed
Heroku fails to build Storybook. The underlying `yarn storybook:build` command fails locally too. I believe this is related to something that I changed in `v7` 🤔 (potentially during dependency deduping?). I didn't spend too much time digging into why the theoretically supported configuration in `config.js` no longer worked.

It appears that this config change allows the command to succeed and has no regressions in `yarn storybook`.

Build error: https://share.getcloudapp.com/9Zuom1qQ

## QA
- [x] Run `yarn storybook:build` confirm that it works as expected (edit: deployed to Heroku)
- [ ] Regression test stories at https://lp-components.herokuapp.com/


chawes13:
@mwislek I deployed production off of this branch to ensure that it not only worked locally, but also in a hosted environment: https://lp-components.herokuapp.com/

.storybook/main.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  stories: ['../**/*.story.js'],
+  addons: [
```
mwislek, line 3: Wow. Strange that you had to do this.

mwislek (approved):
Looks good. I'm sure you noticed the warnings:
```
info => Using implicit CSS loaders
(node:87882) DeprecationWarning: Default PostCSS plugins are deprecated. When switching to '@storybook/addon-postcss',
you will need to add your own plugins, such as 'postcss-flexbugs-fixes' and 'autoprefixer'.

See https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-default-postcss-plugins for details.
(Use `node --trace-deprecation ...` to show where the warning was created)
```
We should probably create an issue to fix this. In any case though, `yarn storybook` and `yarn build:storybook` seem to work fine.

PR #580, opened 2022-12-05
chore: upgrade Node and all minor dependencies

mwislek:
- Upgrade all dependency minor versions
- Upgrade Node version to ^18.12.1

This work is being done as part of a client-template upgrade comparable to the update done for the Ionic client template.
## Author Checklist

- [ ] Add unit test(s)
- [x] Assign dev reviewer


.travis.yml
```diff
@@ -1,4 +1,5 @@
 language: node_js
+dist: focal
```
mwislek, line 2: See: https://travis-ci.community/t/the-command-npm-config-set-spin-false-failed-and-exited-with-1-during/12909/6 and https://github.com/nodejs/node/issues/42351#issuecomment-1068424442. The latest version of Node requires an upgraded Linux distribution, which apparently must be specified within the Travis build. Note, however, that this fix still doesn't build because of an issue with node-sass. I'm going to try upgrading to `sass` (which is recommended anyway) and the package.json `build:styles` script to use `sass`.

PR #584, opened 2023-02-02
Filter invalid span props

jhp0621:
## Author Checklist

- [ ] Add unit test(s)
- [ ] Assign dev reviewer

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,8 @@
     "cleave.js": "^1.6.0",
     "date-fns": "^2.28.0",
     "filter-invalid-dom-props": "^2.1.0",
+    "html-attributes": "^1.1.0",
+    "html-element-attributes": "^3.1.0",
     "lodash": "^4.17.21",
     "prop-types": "^15.8.1",
     "react-color": "^2.19.3",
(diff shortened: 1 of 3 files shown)


jhp0621:
@chawes13 Can I get general feedback on the approach/util before I proceed with adding unit tests and updating other instances of `filterInvalidDOMProps`?

src/utils/local/filter-invalid-html-props.js
```diff
@@ -0,0 +1,54 @@
...
+
+function filterInvalidHtmlProps(element, props) {
+  // This util (filterInvalidDOMProps) covers event handlers such as "onClick" and props that begin with "data-" and "aria-"
+  // but it does not filter out HTML attributes that are not allowed for the element in question
+  const validDOMProps = filterInvalidDOMProps(props)
+
+  const notAllowedAttributes = generateNotAllowedAttributes(element)
+  return omitBy(validDOMProps, (_, key) =>
+    notAllowedAttributes.includes(htmlAttributes[key])
+  )
```
chawes13, line 41: This should _probably_ be optimized, but I think we should only look at that after benchmarking `filterInvalidHtmlProps` against `filterinvalidDOMProps`.

It probably does make sense to write specs first, since then you can change the underlying code under green with confidence.

Issue #567, opened 2022-06-30
Replace Enzyme with React Testing Library


chawes13:
This will accommodate allowing this library to be confidently tested against React 18.

Enzyme does not officially support v18 (or v17 for that matter). The author of the unofficial v17 adapter has recommended switching to RTL and considers Enzyme "dead".

https://github.com/enzymejs/enzyme/issues/2524
https://dev.to/wojtekmaj/enzyme-is-dead-now-what-ekl
https://testing-library.com/docs/react-testing-library/migrate-from-enzyme/

josiasds:
@chawes13 I assume we can close this issue now?

chawes13:
😲 I guess we didn't use the right keywords to link this issue! Yes, definitely.

PR #586, opened 2023-03-15
Migrate enzyme to testing library

josiasds:
This PR contains the migration from `enzyme` to `react-testing-library`, as requested in #567.

From [the docs](https://testing-library.com/docs/react-testing-library/migrate-from-enzyme):

> Testing Library has a different approach to testing when compared to Enzyme. What that means is there's not a one-to-one mapping of Enzyme features to React Testing Library features.
> 
> React Testing Library encourages you to avoid testing [implementation details](https://kentcdodds.com/blog/testing-implementation-details) like the internals of a component you're testing (though it's still possible). [The Guiding Principles](https://testing-library.com/docs/guiding-principles) of this library emphasize a focus on tests that closely resemble how your web pages are interacted by the users.
> 
> You may want to avoid the following implementation details:
> - Internal state of a component
> - Internal methods of a component
> - Lifecycle methods of a component
> - Child components

## Author Checklist

- [x] Setup RTL side-by-side with Enzyme in order to migrate components progressively
- [x] Migrate ColorPicker (depends on https://github.com/LaunchPadLab/lp-components/pull/596)
- [x] Incorporate code review feedback on feature branch
- [x] Merge PRs containing subsets of migrated tests
- [x] Remove `Enzyme` (and adapter) libs and setup

**Components to migrate**:

- [x] `/controls` (#602)
  - [x] `paginator`
  - [x] `color-picker` (depends on #596)
  - [x] `tab-bar`
- [x] `/forms`
  - [x] `/buttons` (in review in this PR #586)
    - [x] `button-area`
    - [x] `button`
    - [x] `submit-button`
  - [x] `/labels` (#607)
    - [x] `error-label`
    - [x] `input-error`
    - [x] `input-label`
    - [x] `labeled-field`
  - [x] `/inputs`
    - [x] `cloudinary-uploader`
    - [x] `checkbox-group`
    - [x] `checkbox`
    - [x] `cloudinary-file-input` (#590)
    - [x] `color-input`
    - [x] `date-input`
    - [x] `dropdown-checkbox-group`
    - [x] `file-input` (#590)
    - [x] `hidden-input`
    - [x] `icon-input`
    - [x] `input`
    - [x] `masked-input`
    - [x] `radio-group`
    - [x] `range-input`
    - [x] `select`
    - [x] `switch`
    - [x] `textarea`
- [x] `/indicators` (#600)
  - [x] `flash-message-container`
  - [x] `flash-message`
  - [x] `loading-container`
  - [x] `spinner` (in review in this PR)
- [x] `/tables` (#603)
  - [x] `sortable-table`
  - [x] `table-column`
  - [x] `table`
- [x] `modal` (#601)
---

- [x] Add unit test(s)
- [x] Assign dev reviewer


test/forms/buttons/button.test.js
```diff
@@ -1,136 +1,145 @@
...
+  fireEvent.click(button)
 
   expect(onClick).not.toHaveBeenCalled()
 })
 
 test('Button adds variant string to class', () => {
-  const wrapper = shallow(<Button variant="custom"> Hi</Button>)
-  expect(wrapper.hasClass('button-custom')).toBe(true)
+  render(<Button variant="custom"> Hi</Button>)
+  expect(screen.getByRole('button')).toBeInTheDocument()
```
josiasds, line 93: Same thing about this already being covered by the next line.

test/forms/buttons/button.test.js
```diff
@@ -1,136 +1,145 @@
...
+  expect(screen.getByRole('button').getAttribute("class")).toContain("custom")
```
josiasds, line 94: In this case, it might be safer to check for the full class name (`button-custom`). Devs might be relying upon the `.button-custom` class in their CSS.

```suggestion
  expect(screen.getByRole('button')).toHaveClass('button-custom')
```

test/forms/buttons/button.test.js
```diff
@@ -1,136 +1,145 @@
...
 })
 
 test('Button adds type to button', () => {
-  const wrapper = shallow(<Button type="reset"> Hi</Button>)
-  expect(wrapper.props().type).toBe('reset')
+  render(<Button type="reset"> Hi</Button>)
+  expect(screen.getByRole('button').getAttribute("type")).toContain("reset")
```
josiasds, line 99: You can take a look at [jest-dom docs](https://github.com/testing-library/jest-dom#table-of-contents) for some custom matchers like `toHaveAttribute` and `toHaveClass` 🙂

test/forms/buttons/button.test.js
```diff
@@ -1,136 +1,145 @@
...
+  expect(screen.getByRole('button').getAttribute("class")).toContain("in-progress")
+  expect(screen.getByRole('button').getAttribute("class")).toContain("button-large")
 })
 
 test('Button can receive object style prop', () => {
-  const wrapper = shallow(<Button style={{ display: 'none' }}>Submit</Button>)
-  expect(wrapper.find('button').prop('style').display).toEqual('none')
-})
-
-test('Button passes down forwardedRef to button', () => {
```
josiasds, line 128: An alternative to this test would be to use `ref` and test the result:

```
test('Button passes down forwardedRef to button', () => {
  const ref = createRef()
  render(
    <Button id="my-button" ref={ref}>
      Click Me
    </Button>
  )

  act(() => ref.current.focus())

  expect(screen.getByRole('button')).toHaveFocus()
})
```

josiasds (commented):
@aojin Great work on this! I left some comments as we discussed in the call.

test/forms/buttons/button-area.test.js
```diff
@@ -1,18 +1,23 @@
...
 
 test('ButtonArea merges classes correctly', () => {
-  const wrapper = shallow(<ButtonArea className="extra classes" />)
-  expect(wrapper.find('div').hasClass('button-area extra classes')).toBe(true)
+  render(
+    <ButtonArea>
+      <button>Hi</button>
+    </ButtonArea>
+  )
+  expect(screen.getByRole('button').closest('div')).toHaveClass('extra classes')
```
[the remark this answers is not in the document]
aojin: Updated!

package.json
```diff
@@ -46,6 +46,7 @@
     "date-fns": "^2.28.0",
     "filter-invalid-dom-props": "^2.1.0",
     "lodash": "^4.17.21",
+    "moment": "^2.29.4",
```
josiasds, line 49: @aojin `moment` is deprecated. You can use `date-fns`, which is already imported in this project.

chawes13: @josiasds Did y'all make this change?

test/forms/inputs/checkbox-group.test.js
```diff
@@ -96,7 +124,12 @@ test('CheckboxGroup passes down props to children', () => {
...
-  expect(wrapper.find('input.custom-group-class').exists()).toBe(false)
-  expect(wrapper.find('input.custom-input-class').exists()).toBe(true)
+  const { container } = render(<CheckboxGroup {...props} />)
+
+  // make sure fieldset has custom-class
+  const checkboxGroup = screen.getByRole('group', { name: 'Test Group' })
+  expect(checkboxGroup).toHaveClass('custom-group-class')
+
+  // make sure there is no class of custom-class within fieldset
+  expect(container.getElementsByClassName('custom-input-class').length).toBe(2)
```
chawes13, line 134: The comment and the assertion don't align. If there are no custom-input-class elements in the fieldset, wouldn't we expect length `toBe(0)`?

chawes13: Again, docs recommend not using `container`. Is there an alternative approach to explore here?

test/forms/inputs/checkbox-group.test.js
```diff
@@ -61,12 +86,9 @@ test("CheckboxGroup has a legend with the group's label (when provided)", () =>
...
     label: 'Checkbox Group',
     meta: {},
-    options: ['TOGGLED_OPTION'],
   }
-  const wrapper = mount(<CheckboxGroup {...props} />)
-  const legend = wrapper.find('fieldset').first().find('legend')
-  expect(legend).toBeTruthy()
-  expect(legend.text()).toEqual('Checkbox Group')
+  render(<CheckboxGroup {...props} />)
+  expect(within(screen.getByRole('group')).getByText('Checkbox Group')).toBeTruthy()
```
chawes13, line 91: What does `within` get us?

test/forms/inputs/checkbox-group.test.js
```diff
@@ -1,41 +1,69 @@
...
-  const newValue = onChange.mock.calls[0][0]
-  expect(newValue).toEqual([])
+test('CheckboxGroup removes value to array when selected option clicked', async () => {
+  // const props = {
+  //   input: {
+  //     name: 'testGroup',
+  //     value: ['2'],
+  //   },
+  //   meta: {},
+  // }
```
chawes13, line 53: Please remove if we aren't using anymore!

chawes13 (commented):
Thanks! Lots of questions – I'm sure that'll peter out as we all get more comfortable with the new library

package.json
```diff
@@ -80,7 +84,7 @@
     "enzyme": "^3.2.0",
     "eslint": "^7.32.0",
     "husky": "^7.0.4",
-    "jest": "^25.1.0",
+    "jest": "27.5.1",
```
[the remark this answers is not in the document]
chawes13: Handled in #596

src/indicators/spinner.js
```diff
@@ -38,6 +38,7 @@ const defaultProps = {
 function Spinner({ className, ...rest }) {
   return (
     <div
+      role="progressbar"
```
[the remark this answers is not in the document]
josiasds: This is an open question and there seems to be no consensus on how to represent a loading component. I'm considering using `aria-busy="true"` alongside `role="progressbar"`, as MDN says in the [Best Practices](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/progressbar_role) section (though it's not exactly the same in our case).

Here are some links for discussion: 
- https://stackoverflow.com/questions/38704467/how-to-label-a-loading-animation-for-wai-aria
- https://docs-button-fix--fundamenta-temp.netlify.app/components/loading-spinner.html

That said, this component was meant to be a starting point for this PR. I need to revisit its tests.

package.json
```diff
@@ -29,7 +29,7 @@
     "prepare": "husky install",
     "prepublishOnly": "yarn run clean && yarn run build",
     "storybook": "yarn && start-storybook -p 6006",
-    "test": "jest --coverage",
+    "test": "jest --coverage=false",
```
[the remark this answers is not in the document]
josiasds: We temporarily disabled this locally. It's not supposed to be here 🙈

package.json
josiasds: I'm going to remove this and use `date-fns`.

test/forms/inputs/checkbox-group.test.js
```diff
@@ -79,8 +101,14 @@ test('CheckboxGroup does not pass class to children', () => {
...
-  const wrapper = mount(<CheckboxGroup {...props} />)
-  expect(wrapper.find('.custom-class').hostNodes().length).toBe(1)
+  const { container } = render(<CheckboxGroup {...props} />)
+
+  // make sure fieldset has custom-class
+  const checkboxGroup = screen.getByRole('group', { name: 'Test Group' })
+  expect(checkboxGroup).toHaveClass('custom-class')
+
+  // make sure there is no class of custom-class within fieldset
+  expect(container.getElementsByClassName('custom-class').length).toBe(1)
```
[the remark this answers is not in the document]
josiasds: Fixed.

test/forms/inputs/checkbox-group.test.js
josiasds: Updated.

test/forms/inputs/checkbox-group.test.js
josiasds: In this case, nothing. I removed it.

test/forms/inputs/checkbox.test.js
```diff
@@ -1,8 +1,35 @@
...
+    },
+    meta: {},
+  }
+
+  return <Checkbox {...props} />
+}
+
+test('Checkbox is checked with true value', async () => {
+
+  render(<WrappedCheckbox />)
```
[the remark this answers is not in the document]
josiasds: No, we don't. Updated.

test/forms/inputs/checkbox.test.js
```diff
@@ -12,10 +39,31 @@ test('Checkbox toggles value when clicked', () => {
...
-  const newValue = onChange.mock.calls[0][0]
-  expect(newValue).toEqual(true)
+
+  render(<Checkbox {...props} />)
+  const checkbox = screen.getByRole('checkbox')
+
+  expect(checkbox).not.toBeChecked()
+})
+
+test('Checkbox fires onChange when clicked', async () => {
```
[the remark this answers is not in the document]
josiasds: I agree. See `test('Checkbox toggles value when clicked')`.

josiasds (commented):
I updated a bunch of tests addressing comments here. There are a few more to update and then we can merge the pending PRs.

test/forms/inputs/date-input.test.js
```diff
@@ -1,51 +1,88 @@
...
+
+  return <DateInput {...defaultProps} {...props} />
+}
+
 test('DateInput renders the error message when provided', () => {
   const props = { input, meta: { invalid: true, touched: true, error } }
-  const wrapper = mount(<DateInput {...props} />)
-  expect(wrapper.find('.error-message').text()).toBe(error)
+  const { container } = render(<WrappedDateInput {...props} />)
+  expect(within(container).getByText('input error')).toBeTruthy()
```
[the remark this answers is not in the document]
josiasds: Done.

test/forms/inputs/date-input.test.js
```diff
@@ -1,51 +1,88 @@
...
-  const current = wrapper.find('[aria-current="date"]')
-  expect(current.prop('tabIndex')).toBe(0)
+  render(<DateInput {...props} />)
+  const input = screen.getByRole('textbox')
+
+  const user = userEvent.setup()
+
+  await user.click(input)
+
+  const current = screen.getByRole('option', { current: 'date' })
```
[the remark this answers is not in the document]
josiasds: That's right: https://testing-library.com/docs/queries/byrole#current

test/forms/inputs/date-input.test.js
```diff
@@ -1,51 +1,88 @@
-import React from 'react'
-import { mount } from 'enzyme'
+import React, { useState } from 'react'
 import { DateInput } from '../../../src/'
+import { render, screen } from '@testing-library/react'
+import { within } from '@testing-library/dom'
```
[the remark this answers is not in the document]
josiasds: Removed.

test/forms/inputs/date-input.test.js
```diff
@@ -1,51 +1,88 @@
...
+import userEvent from '@testing-library/user-event'
+import moment from 'moment'
```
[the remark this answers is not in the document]
josiasds: I removed `moment`.

test/forms/inputs/dropdown-checkbox-group.test.js
```diff
@@ -1,39 +1,61 @@
...
-  }
-  const wrapper = mount(<DropdownCheckboxGroup {...props} />)
-  wrapper.find('input').simulate('change')
-  const newValue = onChange.mock.calls[0][0]
-  expect(newValue).toEqual([])
+test('DropdownCheckboxGroup removes value to array when selected option clicked', async () => {
+
+  render(<WrappedDropdownCheckboxGroup value={['1']} />)
+
+  const input = screen.getAllByRole('group')[0]
```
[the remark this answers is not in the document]
chawes13: I think I know the answer to this now – `getByRole` will throw if there's more than 1

test/forms/inputs/color-input.test.js
```diff
@@ -1,32 +1,47 @@
...
+
+  render(<WrappedColorInput />)
+
+  const input = screen.getByRole('textbox')
+
+  const user = userEvent.setup()
+
+  await user.click(input)
+
+  await user.keyboard('{0}{0}{0}{Enter}');
```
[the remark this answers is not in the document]
josiasds: No need for this. We can merge `user.click` and `user.keyboard` into `user.type('000{Enter}')`.

test/forms/inputs/dropdown-checkbox-group.test.js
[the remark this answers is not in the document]
josiasds: That's right.

test/forms/inputs/color-input.test.js
```diff
@@ -1,32 +1,47 @@
...
+test('ColorInput expands dropdown when hex input is focused', async () => {
+  const { container } = render(<WrappedColorInput />)
+
+  const input = screen.getByRole('textbox')
+
+  const user = userEvent.setup()
+
+  await user.click(input)
+
+  expect(within(container).getByText('hex')).toBeTruthy()
```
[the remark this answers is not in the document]
josiasds: In this specific test, `getByText` would throw if not found, so there's no difference in practice.

Since we're testing HTML elements though, I vote for `toBeInTheDocument()` to be the default. The element must be added to the DOM to pass this validation, while the other matchers just test for the existence of the element.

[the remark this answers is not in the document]
josiasds: @chawes13 Also, I opted for adding a "dialog" role to the popover element in the `ColorPicker` component. Does that sound right?

src/indicators/spinner.js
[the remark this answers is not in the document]
chawes13: Hmm, what do you think about Bootstrap's implementation? https://getbootstrap.com/docs/5.3/components/spinners/#border-spinner

test/forms/buttons/button-area.test.js
```diff
@@ -1,18 +1,25 @@
...
-  const wrapper = shallow(
+  render(
     <ButtonArea>
       <button>Hi</button>
     </ButtonArea>
   )
-  expect(wrapper.find('div').hasClass('button-area')).toBe(true)
-  expect(wrapper.find('button').contains('Hi')).toBe(true)
+
+  expect(screen.getByRole('button')).toBeInTheDocument()
```
chawes13, line 12: Is this needed still since we'll have this implicit assertion on line 13?

test/forms/buttons/button.test.js
```diff
@@ -1,136 +1,167 @@
...
-  const wrapper = shallow(
+  render(
     <Button onClick={onClick} {...formProps}>
-      {' '}
       Hi
     </Button>
   )
-  wrapper.find('button').simulate('click')
+
+  const user = userEvent.setup()
```
chawes13, line 34: The docs explicitly tell us to run this _before_ the element is rendered.

> We recommend invoking [userEvent.setup()](https://testing-library.com/docs/user-event/setup) before the component is rendered. This can be done in the test itself, or by using a setup function. We discourage rendering or using any userEvent functions outside of the test itself - e.g. in a before/after hook - for reasons described in ["Avoid Nesting When You're Testing"](https://kentcdodds.com/blog/avoid-nesting-when-youre-testing).

https://testing-library.com/docs/user-event/intro/#writing-tests-with-userevent

test/forms/buttons/button.test.js
```diff
@@ -1,136 +1,167 @@
...
     <Button variant="primary" className="button-large">
       Click Me
     </Button>
   )
-  expect(wrapper.hasClass('button-primary')).toBe(true)
-  expect(wrapper.hasClass('button-large')).toBe(true)
+  const button = screen.getByRole('button')
+
+  expect(button).toHaveClass('button-primary')
+  expect(button).toHaveClass('button-large')
```
chawes13, line 123: `toHaveClass` accepts a list as well

https://github.com/testing-library/jest-dom#tohaveclass
```suggestion
  expect(button).toHaveClass('button-primary', 'button-large')
```

test/forms/buttons/button.test.js
```diff
@@ -1,136 +1,167 @@
...
     <Button className="button-large" invalid>
       Click Me
     </Button>
   )
-  expect(wrapper.hasClass('is-disabled')).toBe(true)
-  expect(wrapper.hasClass('button-large')).toBe(true)
+  const button = screen.getByRole('button')
+
+  expect(button).toHaveClass('is-disabled')
+  expect(button).toHaveClass('button-large')
```
chawes13, line 135:
```suggestion
  expect(button).toHaveClass('is-disabled', 'button-large')
```

test/forms/buttons/button.test.js
```diff
@@ -1,136 +1,167 @@
...
-  expect(wrapper.hasClass('in-progress')).toBe(true)
-  expect(wrapper.hasClass('button-large')).toBe(true)
-})
+  const button = screen.getByRole('button')
 
-test('Button can receive object style prop', () => {
-  const wrapper = shallow(<Button style={{ display: 'none' }}>Submit</Button>)
-  expect(wrapper.find('button').prop('style').display).toEqual('none')
+  expect(button).toHaveClass('in-progress')
+  expect(button).toHaveClass('button-large')
```
chawes13, line 147:
```suggestion
  expect(button).toHaveClass('in-progress', 'button-large')
```

test/forms/inputs/hidden-input.test.js
```diff
@@ -1,29 +1,25 @@
...
-    'visibility',
-    'hidden'
-  )
+  render(<HiddenInput {...props} data-testid="test" />)
+
+  const wrapper = screen.getByTestId('hidden-input-wrapper')
+
+  expect(wrapper).toHaveStyle('visibility: hidden')
+  expect(wrapper).toHaveStyle('left: -9999px')
+  expect(wrapper).toHaveStyle('position: absolute')
```
chawes13, line 24: I'm not a huge fan of this test. Can we replace it with a combination of toBeInTheDocument and [toBeVisible](https://github.com/testing-library/jest-dom#tobevisible)? And in that case would we _need_ to test the div with the testid or could we keep this assertion on the element with the `textbox` role?

test/forms/inputs/dropdown-checkbox-group.test.js
```diff
@@ -1,39 +1,64 @@
...
-  }
-  const wrapper = mount(<DropdownCheckboxGroup {...props} />)
-  wrapper.find('input').simulate('change')
-  const newValue = onChange.mock.calls[0][0]
-  expect(newValue).toEqual([])
+test('DropdownCheckboxGroup removes value from array when selected option clicked', async () => {
+  render(<WrappedDropdownCheckboxGroup value={['1']} />)
+
+  const select = screen.getAllByRole('group')[0]
+  const user = userEvent.setup()
```
chawes13, line 53: Let's move this above `render` (same goes for the rest of the tests in this file)

test/forms/inputs/date-input.test.js
```diff
@@ -1,51 +1,81 @@
...
 })
 
-test('DateInput updates the value on change', () => {
-  const props = { input, meta: {} }
-  const wrapper = mount(<DateInput {...props} />)
-  expect(wrapper.find('input').props().value).toEqual('01/01/2020')
+test('DateInput updates the value on change', async () => {
+  render(<WrappedDateInput />)
+
+  const user = userEvent.setup()
```
chawes13, line 38: Let's move this above `render` in all tests

test/forms/inputs/color-input.test.js
```diff
@@ -1,32 +1,42 @@
...
-  wrapper.find('.hex-input').simulate('change', { target: { value: '000' } })
-  expect(onChange).toHaveBeenCalledWith('#000')
+
+  return <ColorInput {...defaultProps} {...props} />
+}
+
+test('ColorInput hex input adds hash to value', async () => {
+  render(<WrappedColorInput />)
+
+  const user = userEvent.setup()
```
chawes13, line 25: Let's move this above `render` in all tests

test/forms/inputs/color-input.test.js
```diff
@@ -1,32 +1,42 @@
...
-  const props = {
+const WrappedColorInput = (props) => {
+  const [value, setValue] = useState('')
+
+  const defaultProps = {
     input: {
       name: 'test',
-      value: '',
-      onChange,
+      value: value,
```
chawes13, line 13: 🍔 You can use shorthand assignment here (i.e., `value,`)

test/forms/inputs/checkbox.test.js
```diff
@@ -30,8 +79,10 @@ test('Checkbox is given an aria-describedby attribute when there is an input err
       invalid: true,
     },
   }
-  const wrapper = mount(<Checkbox {...props} />)
-  expect(wrapper.find('input').prop('aria-describedby')).toContain(name)
+  render(<Checkbox {...props} />)
+  const checkbox = screen.getByRole('checkbox')
+
+  expect(checkbox.getAttribute('aria-describedby')).toContain(name)
```
chawes13, line 85: Replace with `toHaveAttribute`

test/forms/inputs/checkbox-group.test.js
```diff
@@ -61,12 +75,9 @@ test("CheckboxGroup has a legend with the group's label (when provided)", () =>
...
     label: 'Checkbox Group',
     meta: {},
-    options: ['TOGGLED_OPTION'],
   }
-  const wrapper = mount(<CheckboxGroup {...props} />)
-  const legend = wrapper.find('fieldset').first().find('legend')
-  expect(legend).toBeTruthy()
-  expect(legend.text()).toEqual('Checkbox Group')
+  render(<CheckboxGroup {...props} />)
+  expect(screen.getByText('Checkbox Group')).toBeTruthy()
```
chawes13, line 80: Let's replace with `toBeInTheDocument()`

test/forms/inputs/checkbox-group.test.js
```diff
@@ -45,12 +62,9 @@ test("CheckboxGroup has a legend with the group's name by default", () => {
...
     },
     meta: {},
-    options: ['TOGGLED_OPTION'],
   }
-  const wrapper = mount(<CheckboxGroup {...props} />)
-  const legend = wrapper.find('fieldset').first().find('legend')
-  expect(legend).toBeTruthy()
-  expect(legend.text()).toEqual('Test Group')
+  render(<CheckboxGroup {...props} />)
+  expect(screen.getByText('Test Group')).toBeTruthy()
```
chawes13, line 67: Let's replace with `toBeIntheDocument()`

test/forms/inputs/checkbox-group.test.js
```diff
@@ -1,41 +1,58 @@
...
-  }
-  const wrapper = mount(<CheckboxGroup {...props} />)
-  wrapper.find('input').simulate('change')
-  const newValue = onChange.mock.calls[0][0]
-  expect(newValue).toEqual([])
+test('CheckboxGroup removes value to array when selected option clicked', async () => {
+  render(<WrappedCheckboxGroup />)
+
+  const checkbox2 = screen.getByRole('checkbox', { name: 'Second Option' })
+  const user = userEvent.setup()
```
chawes13, line 48: Let's move above `render` (in this and any other test in this file)

chawes13 (commented):
Looking pretty good! A few small items to clean up, but I think we can start merging in the other PRs if you're comfortable with that (once you've approved them)

test/forms/buttons/button-area.test.js
josiasds: Removed.

test/forms/buttons/button.test.js
josiasds: Oh I didn't know that 👀

test/forms/inputs/hidden-input.test.js
josiasds: Yes, 100%. I removed this test and added the assertion that the input is in the DOM but not visible. It shouldn't matter how we accomplish that in CSS.

src/indicators/spinner.js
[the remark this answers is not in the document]
chawes13: Maybe, but not a part of this PR. Let's stick with progressbar for now

test/forms/inputs/date-input.test.js
```diff
@@ -28,61 +27,56 @@ const WrappedDateInput = (props) => {
...
 
+  const dateInput = screen.getByRole('textbox')
+  await user.click(dateInput)
   const current = screen.getByRole('option', { current: 'date' })
-  const today = moment().format('dddd, MMMM Do, YYYY')
-  const currentDateAriaALabel = screen.getByLabelText(`Choose ${today}`)
-  expect(current).toBeTruthy()
-  expect(currentDateAriaALabel).toBeTruthy()
+
+  expect(current).toHaveProperty('tabIndex', 0)
```
chawes13, line 81: I don't think this assertion adequately covers the scope test case. It confirms that the `current` item has a tab index of 0, but it **does not** confirm that the value is equal to today's date.

src/utils/local/trigger-on-keys.js
```diff
@@ -16,7 +16,7 @@ import { castArray, compact } from 'lodash'
 function triggerOnKeys(fn, keyCodes) {
   const codes = compact(castArray(keyCodes))
   return function (e) {
-    const key = e.which || e.keyCode
+    const key = e.key || e.code
```
chawes13, line 19: The `user.keyboard({Enter})` interaction wasn't passing for this after we switched to using `keydown`. Turns out we were using non-standard & deprecated codes. Switched to what they recommended: https://github.com/testing-library/user-event/issues/842.

This shouldn't be considered a breaking change IMO because `triggerOnKeys` is used internally

chawes13:
@josiasds I merged the latest release on `main` into and migrated color picker tests. LMK what else you need from me!

josiasds:
@chawes13 I can't approve this PR since I'm the author, but it's ready!

chawes13:
Thanks @josiasds! I want to fix the test coverage regression before we officially go-live with this. I'll tackle that first thing when I'm back 🙂

test/forms/helpers/read-files-as-data-urls.test.js
```diff
@@ -0,0 +1,35 @@
...
+
+// HELPERS
+
+function setWindowValue(value) {
+  // eslint-disable-next-line no-undef
+  const originalWindow = global.window
+  // eslint-disable-next-line no-undef
+  Object.defineProperty(global, 'window', {
+    value
+  })
```
chawes13, line 32: Setting `global.window = undefined` directly doesn't work, since I believe it's considered read only. This is a hacky way around it. LMK if you know any other "official" approaches.

I went down the rabbit hole of trying to mock a function within a function (i.e., `isServer`), but there doesn't seem to be a clear way to do that in Jest (without also exporting that local function from the module). 

https://stackoverflow.com/questions/39755439/how-to-mock-imported-named-function-in-jest-when-module-is-unmocked

test/forms/helpers/read-files-as-data-urls.test.js
```diff
@@ -0,0 +1,35 @@
+import { readFilesAsDataUrls } from '../../../src'
+
+describe('readFilesAsDataUrls', () => {
```
chawes13, line 3: I believe the switch from `node` to `jsdom` is what flipped the 66% to 33%. We did however change the implementation so that `readFiles` should only ever be called after user input (on change). While the `isServer()` guard is probably no longer needed, I'd like to keep this a minor change by writing the test for the functionality as written today, instead of removing that guard.

It's definitely a confusing topic so let me know if you have any questions!

test/forms/helpers/read-files-as-data-urls.test.js
chawes13: We do have a handful of components with this guard in place, so it might make sense to spend time in the future investigating how to do this "properly".

chawes13: Nevermind, I think I figured out a better solution!

src/tables/helpers/get-column-data.js
```diff
@@ -1,7 +1,7 @@
 import { castArray, has } from '../../utils'
 
 // Get column info from children via props
-function getColumnData(children = [], doDisable) {
+function getColumnData(children, doDisable) {
```
chawes13, line 4: This was only called in `sortable-table`, whose prop types **require** children. We were getting a flag for missing a branch here, but in reality this isn't a valid use of this util.

josiasds:
@chawes13 I just reviewed your latest changes, looking good. Mocking `isServer` is the best approach in my opinion.

src/controls/tab-bar/focus.js
```diff
@@ -1,33 +1,40 @@
...
-// Funnction that can be passed to event handlers (e.g., onKeyPress) that manages which element should be focused
+// Function that can be passed to event handlers (e.g., onKeyPress) that manages which element should be focused
 // Note: Expected keyboard interaction with arrow keys changes depending on the orientation of the tab list
-function manageFocus(e, { vertical = false }) {
+function manageFocus(e, { vertical }) {
   // If not activated while on a tab, then ignore
   if (!isTabControl(e.target)) return
 
-  const key = (e.which || e.keyCode || '').toString()
+  const key = e.key === 'Unidentified' ? e.code : e.key
```
chawes13, line 16: If `key` cannot be identified, it returns `Unidentified`

Source: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key#value

src/controls/tab-bar/focus.js
```diff
@@ -81,6 +88,7 @@ function getAdjacentControl(control, { previous = false } = {}) {
 
 // Recursively searches for the closest parent tab list
 function getClosestTabList(el) {
+  /* istanbul ignore next */
```
chawes13, line 91: This isn't possible to test without changing the markup of the component. If this situation does occur, we'll get an error on `undefined.querySelectorAll('[role="tab"]'))`.

However, it still feels good to me to have this guard in a recursive function? Let me know if you think we should remove it outright

src/forms/inputs/cloudinary-file-input/cloudinary-uploader.js
```diff
@@ -78,7 +78,7 @@ const DEFAULT_REQUEST_OPTIONS = {
 const FILE_NAME_PATTERN = /[\s?&#\\%<>]/gi
 
 // Throws an error when a required param is not found
-function requireParam(paramName, context = 'Error') {
+function requireParam(paramName, context) {
```
chawes13, line 81: This function is never used without passing `context`, so we don't need this default

src/controls/tab-bar/focus.js
josiasds: TIL 😃
Would it help to have a comment in the code?

chawes13:
@josiasds Ready for ya! We got to 100% coverage 😎

test/forms/inputs/dropdown-checkbox-group.test.js
```diff
@@ -62,3 +61,19 @@ test('DropdownCheckboxGroup removes value from array when selected option clicke
...
   expect(selectValueLabel).toBeInTheDocument()
 })
+
+test('DropdownCheckboxGroup sets menu no longer active when clicked outside', async () => {
+  const user = userEvent.setup()
+
+  render(<WrappedDropdownCheckboxGroup value={['1']} />)
+
+  const select = screen.getByRole('button')
+  await act(() => user.click(select))
```
josiasds, line 71: Why is `act` necessary here?

chawes13: My understanding is that any code that relies on asserting on state changes within a component should be wrapped in `act`. The button's `nextSibling` will only have the `is-active` class applied if the internal `expanded` state changes

chawes13: Hmm, it looks like all events within React's callstack should be covered, so maybe I don't need it here!

https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning

Additionally, it's possible that instead of calling `act` directly, I should be using `waitFor` (https://testing-library.com/docs/react-testing-library/cheatsheet/#async). What do you think?

chawes13: Ok, fixed it :)

test/controls/tab-bar.test.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react'
-import { render, screen } from '@testing-library/react'
+import { render, screen, fireEvent } from '@testing-library/react'
```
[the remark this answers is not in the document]
chawes13: Yes, from what I could tell the `keyboard` command only works on focusable elements and there are none _other_ than the tabs themselves. Most of the tests that use `fireEvent` needed an event to be targeted at an element that was _not_ a tab (e.g., pressing directional key when not focused on a tab does not change focus). Lastly, one of the tests required `key` to be Unidentified, which I wasn't able to replicate with any known keyboard commands (so I used `{ code: 'End' }` with `fireEvent` instead).

It's a little bit of an escape hatch – I used `keyboard` when possible.

PR #589, opened 2023-04-26
Enable passage of props when using custom `rowComponent` with SortableTable/Table

vanderhoop:
At the moment, the `SortableTable` and `Table` components allow the passage of a custom `rowComponent`. However, there's no ability to pass additional props to this component from parent/consuming components, as this is presumed to just need the `data`.

In my current project, we're building a rather involved table, and it requires the passage of event handlers to a `rowComponent` from a parent. At the moment, our workaround is to either **a)** create the component in a closure that has access to the event handler, or **b)** force us to manage the state in a grander fashion with either contexts or redux. Neither option is ideal, as we're only passing the handler down through one component, and it occurred to us that this is something that the `lp-components` package could introduce at a very low complexity cost.

## Release Notes

Semver minor bump (new behavior without breaking existing contracts), adding `rowComponentProps`, that get passed to the row component.

## Author Checklist

- [x] Add unit test(s)
- [x] Assign dev reviewer

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -1410,6 +1410,7 @@ This component's behavior is largely determined by the [TableColumn][103] compon
 *   `controlled` **[Boolean][144]** A flag to disable sorting on all columns, while keeping the sorting arrows. Used when sorting is controlled by an external source. (optional, default `false`)
 *   `onChange` **[Function][143]?** A callback that will be fired when the sorting state changes
 *   `rowComponent` **[Function][143]?** A custom row component for the table. Will be passed the `data` for the row, several internal table states (the current column being sorted (sortPath), whether ascending sort is active or not (ascending), the sorting function (sortFunc), and the value getter (valueGetter)) as well as `children` to render.
+*   `rowComponentProps` **[Object][148]?** Additional props to pass to the rowComponent, useful for passing event handlers
 *   `headerComponent` **[Function][143]?** A custom header component for the table. Will be passed the configuration of the corresponding column, as well as the current `sortPath` / `ascending` and an `onClick` handler. May be overridden by a custom `headerComponent` for a column.
 
 ### Examples
--- a/stories/tables/sortable-table.story.js
+++ b/stories/tables/sortable-table.story.js
@@ -31,8 +31,8 @@ function CustomCellWithRowData({ data: { name, active } }) {
 }
 
 // eslint-disable-next-line react/prop-types
-function CustomRow({ data: { active }, children }) {
-  return <tr style={{ backgroundColor: colorForStatus(active) }}>{children}</tr>
+function CustomRow({ data: { active }, children, onClick }) {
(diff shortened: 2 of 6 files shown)


stories/tables/sortable-table.story.js
```diff
@@ -114,8 +114,8 @@ storiesOf('SortableTable', module)
       <Column name="active" component={CustomCellWithRowData} />
     </SortableTable>
   ))
-  .add('with custom row component', () => (
-    <SortableTable data={tableData} rowComponent={CustomRow}>
+  .add('with custom row component and props', () => (
+    <SortableTable data={tableData} rowComponent={CustomRow} rowComponentProps={{onClick: () => { console.log('Clicked') }}}>
```
vanderhoop, line 118: I thought about getting more involved with what this handler did, but I didn't want to invest too much time, as I didn't know if the high-level proposal would be ❤️ 'd. Can update if reviewers feel it's necessary. Lemmeknow!

chawes13:
@vanderhoop Is `onClick` the only event handler that you have a use case for in your client project? Or are there other dynamic handlers/attributes that the `tr` requires?

stories/tables/sortable-table.story.js
chawes13: As far as the story goes, I'd like the event handler to reference dynamic data. As implemented, this could be done exclusively through a custom `rowComponent`.

Let's hold off _briefly_ on making this change while we connect on https://github.com/LaunchPadLab/lp-components/pull/589#issuecomment-1559491654.

docs.md
```diff
@@ -1410,6 +1410,7 @@ This component's behavior is largely determined by the [TableColumn][103] compon
 *   `controlled` **[Boolean][144]** A flag to disable sorting on all columns, while keeping the sorting arrows. Used when sorting is controlled by an external source. (optional, default `false`)
 *   `onChange` **[Function][143]?** A callback that will be fired when the sorting state changes
 *   `rowComponent` **[Function][143]?** A custom row component for the table. Will be passed the `data` for the row, several internal table states (the current column being sorted (sortPath), whether ascending sort is active or not (ascending), the sorting function (sortFunc), and the value getter (valueGetter)) as well as `children` to render.
+*   `rowComponentProps` **[Object][148]?** Additional props to pass to the rowComponent, useful for passing event handlers
```
chawes13, line 1413: This is _bit_ pedantic, but I'd prefer to call this `rowProps`. Otherwise, it makes me believe that I **have** to specify a `rowComponent` in order to receive those props, which isn't the case in the current implementation.

vanderhoop:
hey @chawes13, thanks for taking a look at this. We no longer need this functionality, as we introduced actual `button` elements into the cells that need them, which is better for accessibility than clickable `trs` that lack keyboard event listeners. As such, I'm going to close this, lest it do harm!

PR #592, opened 2023-06-21
Remove Recompose

### LINKED ISSUE #585: Remove recompose
## Feature Description
The [original project](https://github.com/acdlite/recompose) is archived. While there is a [community fork](https://github.com/react-recompose/react-recompose), we've removed it from the client-template and we're only using the `wrapDisplayName` util in a story, so I don't see much value in porting over to that library.

## Suggested Solution
Rewrite the `wrapDisplayName` util locally so that it can still be imported by client apps

danparnella:
## Release Notes
- Removed the recompose library and pulled in local version of `wrapDisplayName` so that components and external imports would be backwards compatible
- Resolves #585 

## Author Checklist

- [x] Add unit test(s)
- [x] Assign dev reviewer


migration-guides/v9.0.0.md
```diff
@@ -0,0 +1,22 @@
...
+
+While this won't break any imported components (they have been updated to work without recompose), it will break any instance of directly importing `wrapDisplayName` from the `utils` folder.
+
+To resolve this, remove that import and replace
+```
+wrapDisplayName(<component>, 'stringWrapper')
+``` 
+with an interpolated string:
+```
+`stringWrapper(${<component>.name})`
```
chawes13, line 21: Hmm, this isn't an _exact_ 1:1 replacement of the util. The first property it checks is `displayName`, followed by `name` ([source](https://github.com/acdlite/recompose/blob/3db12ce7121a050b533476958ff3d66ded1c4bb8/src/packages/recompose/getDisplayName.js#LL10C3-L10C64)).

Are any of the usages in this library where `<component>.displayName` yields a different, truthy result?

danparnella: Thanks, yeah I definitely should have just gone straight to the source!  Would it be worth it to actually just pull in the whole `wrapDisplayName` as a local util? It's clearly not that complicated and then this might not actually be a breaking change as `wrapDisplayName` would still exist in `utils`.

chawes13: Yep, let's do it. Maybe you can steal their tests too 🤪

src/utils/local/wrap-display-name.js
```diff
@@ -0,0 +1,21 @@
...
+  if (typeof Component === 'string') {
+    return Component
+  }
+
+  if (!Component) {
+    return undefined
+  }
+
+  return Component.displayName || Component.name || 'Component'
+}
```
danparnella, line 19: I didn't pull this out separately since we weren't using it separately originally.

test/utils/wrap-display-name.test.js
```diff
@@ -0,0 +1,62 @@
...
+    class SomeComponent extends React.Component {
+      render() {
+        return <div />
+      }
+    }
+
+    expect(wrapDisplayName(SomeComponent, 'someHoC')).toBe(
+      'someHoC(SomeComponent)'
+    )
+  })
```
danparnella, line 15: This was actually the only test they had lol. Added a few more!

stories/dynamic-input.js
```diff
@@ -1,9 +1,5 @@
 import React, { Component } from 'react'
-import wrapDisplayName from 'recompose/wrapDisplayName'
-import set from 'lodash/fp/set'
-import get from 'lodash/fp/get'
-import noop from 'lodash/noop'
-import compose from 'lodash/fp/compose'
+import { wrapDisplayName, set, get, noop, compose } from '../src/utils'
```
chawes13, line 2: 👍

test/utils/wrap-display-name.test.js
```diff
@@ -0,0 +1,62 @@
+import React from 'react'
+import { wrapDisplayName } from '../../src/utils'
+
+describe('Wrap Display Name', () => {
+  test('wraps the display name of a React component with the name of an HoC', () => {
```
chawes13, line 5: A little pedantic but I think omitting "display" here will make this test more clearly stand out in comparison to the next test case

chawes13 (approved):
Nice work 🚀 ! One non-blocking comment for ya

PR #595, opened 2023-08-01
Labeled field updates

jhp0621:
Addresses: 
https://github.com/LaunchPadLab/lp-components/issues/436
https://github.com/LaunchPadLab/lp-components/issues/624

## Author Checklist

- [x] Add unit test(s)
- [x] Assign dev reviewer


src/forms/inputs/checkbox-group.js
```diff
@@ -88,17 +91,45 @@ const propTypes = {
...
-  return <legend>{label || convertNameToLabel(name)}</legend>
+function CheckboxGroupLegend({
+  label,
+  name,
+  required,
+  requiredIndicator,
+  hint,
+}) {
+  return (
+    <legend className={classnames({ 'visually-hidden': label === false })}>
```
jhp0621, line 112: source: https://www.w3.org/TR/WCAG20-TECHS/H71.html#:~:text=The%20first%20element%20inside%20the,related%20radio%20buttons%20and%20checkboxes.
> The first element inside the fieldset must be a legend element, which provides a label or description for the group.

src/forms/labels/labeled-field.js
```diff
@@ -88,23 +91,25 @@ function LabeledField({
...
   const { touched, invalid } = meta
   return (
-    <fieldset
-      className={classnames(className, {
+    <Component
+      className={classnames(className, 'field-wrapper', {
         error: hasInputError({ touched, invalid }),
         disabled: rest.disabled,
       })}
+      role="group"
```
jhp0621, line 105: **aria role = group** MDN: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/group_role

Does it make sense to assign `group` role to the container here? 

It is technically redundant when Component is `fieldset` whose implicit aria role is group (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset#technical_summary)

This would be for when Component is `div`:
Screen Shot 2023-09-25 at 11 39 51 AM
This role assignment would be necessary to pass the test that relies on role (if we are to keep the test as is): https://github.com/LaunchPadLab/lp-components/blob/main/test/forms/inputs/icon-input.test.js#L12

What about for individual checkbox/radio group? Is it odd that there is a `role=group` for the container that wraps the individual label and the corresponding checkbox?
Screen Shot 2023-09-25 at 11 38 41 AM

cc: @chawes13

danparnella (commented):
A couple quick thoughts. Will try to add more tonight or tomorrow morning.

src/forms/inputs/checkbox-group.js
```diff
@@ -123,30 +144,41 @@ function CheckboxGroup(props) {
       return onChange(newValueArray)
     }
   }
+
+  const CheckboxOptions = () =>
+    optionObjects.map((option, i) => {
+      return (
+        <Checkbox // eslint-disable-line react/jsx-key
```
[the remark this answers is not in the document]
chawes13: We should also not pass `i` for key 🙂. What property do we have on `option` that must be unique?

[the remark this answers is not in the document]
danparnella: Ji already updated that to `option.value`. It was in my original comment and I removed that before submitting when I saw the recent changes.

src/forms/inputs/checkbox-group.js
```diff
@@ -34,6 +36,7 @@ import {
  * @param {Object} meta - A `redux-form` [meta](http://redux-form.com/6.5.0/docs/api/Field.md/#meta-props) object
  * @param {Array} options - An array of checkbox values (strings, numbers, or key-value pairs)
  * @param {Object} [checkboxInputProps={}] - An object of key-value pairs representing props to pass down to all checkbox inputs
+ * @param {Boolean} [useDropdown] - A boolean that determines whether the checkbox options are displayed in a dropdown container or not (optional, default `false`)
```
danparnella, line 39: I think if you update `[useDropdown]` to `[useDropdown=false]`, that will automatically add in the "(optional, default `false`)" into the generated Docs (like `checkboxInputProps` above).

src/forms/inputs/checkbox-group.js
```diff
@@ -88,17 +91,45 @@ const propTypes = {
...
+        <span className="required-indicator" aria-hidden="true">
+          {requiredIndicator}
+        </span>
+      )}
+      {hint && <i>{hint}</i>}
+    </legend>
+  )
+}
+
+function CheckboxOptionContainer({ children, useDropdown, value }) {
```
danparnella, line 124: Nit: Should we pluralize `Option` in `CheckboxOptionContainer` since this wraps all of the options?

danparnella (changes requested):
Still going through it, but here's a bit more feedback!

src/forms/inputs/checkbox-group.js
chawes13, line 39: I'm pretty hesitant on the naming here. `useX` in React-land now has a very specific set of expectations with it after the release of hooks. I'd prefer to leave that prefix for custom hooks and not for prop names, if possible.

src/forms/inputs/radio-group.js
```diff
@@ -89,9 +90,18 @@ const defaultProps = {
   radioInputProps: {},
 }
 
-function RadioGroupLegend({ label, name }) {
-  if (label === false) return null
-  return <legend>{label || convertNameToLabel(name)}</legend>
+function RadioGroupLegend({ label, name, required, requiredIndicator, hint }) {
+  return (
+    <legend className={classnames({ 'visually-hidden': label === false })}>
```
chawes13, line 95: I think we'll also need to add a note about this in the migration guide. If the user is coming from an older project, they might not have `visually-hidden` styles included in their scss set up (or they might have it applied by another name).

src/forms/labels/labeled-field.js
```diff
@@ -7,9 +7,9 @@ import { hasInputError } from '../helpers'
 
 /**
  *
- * A fieldset wrapper for redux-form controlled inputs. This wrapper adds a label component (defaults to {@link InputLabel})
+ * A container wrapper for redux-form controlled inputs. This wrapper adds a label component (defaults to {@link InputLabel})
```
chawes13, line 10: IMO "container wrapper" is redundant. It's a wrapper or a container, but not both?

src/forms/labels/labeled-field.js
```diff
@@ -20,6 +20,7 @@ import { hasInputError } from '../helpers'
  * @param {Boolean} [hideErrorLabel] - A boolean determining whether to hide the error label on input error (optional, default `false`)
  * @param {Function} [labelComponent=InputLabel] - A custom label component for the input
  * @param {Function} [errorComponent=InputError] - A custom error component for the input
+ * @param {Element Type} [as] - A string that determines the element type of the component (optional, default `div`)
```
chawes13, line 23: You can specify props as optional and their default automatically using JSDoc syntax. Refer to `labelComponent` above for how to accomplish that. While we're here, we might as well make that update to `hideErrorLabel` as well.

chawes13, line 23: If this prop was really a string, then the @param should be `String` right? But `propTypes.elementType` supports functional components and class components as well (https://github.com/facebook/prop-types/pull/211/files).

Do you have a source for using `elementType` for `"as"`?

docs.md
```diff
@@ -1112,19 +1065,19 @@ function EmailInput ({
...
+use the [omitLabelProps][88] helper.
 
 ### Parameters
 
-*   `hideErrorLabel` **[Boolean][144]?** A boolean determining whether to hide the error label on input error (optional, default `false`)
-*   `labelComponent` **[Function][143]** A custom label component for the input (optional, default `InputLabel`)
-*   `errorComponent` **[Function][143]** A custom error component for the input (optional, default `InputError`)
+*   `hideErrorLabel` **[Boolean][141]?** A boolean determining whether to hide the error label on input error (optional, default `false`)
+*   `labelComponent` **[Function][140]** A custom label component for the input (optional, default `InputLabel`)
+*   `errorComponent` **[Function][140]** A custom error component for the input (optional, default `InputError`)
```
chawes13, line 1080: Hmm, I would expect `as` to show up here. Maybe it's not showing up because JSDoc doesn't respect "Element Type" as a type? I don't think we can assume the naming convention in the prop types library maps 1:1 to JSDoc

stories/forms/inputs/checkbox-group.story.js
```diff
@@ -75,3 +75,6 @@ storiesOf('CheckboxGroup', module)
       }}
     />
   ))
+  .add('with dropdown options', () => (
+    <CheckboxGroup input={inputProps} meta={{}} options={options} useDropdown />
```
chawes13, line 79: Prefer to be explicit here, based on rationale presented here: https://legacy.reactjs.org/docs/jsx-in-depth.html#props-default-to-true

stories/forms/inputs/checkbox-group.story.js
```diff
@@ -75,3 +75,6 @@ storiesOf('CheckboxGroup', module)
       }}
     />
   ))
+  .add('with dropdown options', () => (
```
chawes13, line 78: The name of this story almost reads as if the options themselves have some form of dropdown capability. Could you rephrase this to be more clear?

chawes13 (commented):
In addition to adding a migration guide that details the changes that a user would have to make (including updating their stylesheets), we should address the regression in test coverage. Right now on main we're very close to 100%, this PR leaves some pathways uncovered (report available via `yarn test`).

**Main**


**This branch**

.storybook/styles/application.scss
```diff
@@ -48,10 +48,9 @@ $output-bourbon-deprecation-warnings: false;
 @import 'components/progress-circle';
 @import 'components/login';
 @import 'components/pagination';
-@import 'components/sidebar-content';
```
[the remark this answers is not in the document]
danparnella: Ji was referencing my original PR where I removed a couple stylesheets that weren't being used: https://github.com/LaunchPadLab/lp-components/pull/514/files#r1132970196.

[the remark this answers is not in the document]
chawes13: Thanks. Are we good to close that PR in favor of this one?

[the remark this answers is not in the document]
danparnella: Closed! I'm still reviewing them side-by-side to see if there's anything else we can pull from there, but it was safe to close.

src/forms/inputs/radio-group.js
```diff
@@ -89,9 +90,18 @@ const defaultProps = {
   radioInputProps: {},
 }
 
-function RadioGroupLegend({ label, name }) {
-  if (label === false) return null
-  return <legend>{label || convertNameToLabel(name)}</legend>
+function RadioGroupLegend({ label, name, required, requiredIndicator, hint }) {
```
danparnella, line 93: Since this is the same for the checkbox group, would it make sense to pull this out into it's own component (just a `legend` component)? That way it could also be used separately on projects if needed.

src/forms/inputs/checkbox-group.js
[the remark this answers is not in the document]
danparnella: Marking unresolved as I think we should also remove the lint disabling by pulling `key` up out of the props spread.

src/forms/inputs/radio-group.js
```diff
@@ -137,6 +147,7 @@ function RadioGroup(props) {
     <LabeledField
       className={className}
       labelComponent={RadioGroupLegend}
+      as="fieldset"
       {...props}
     >
       {optionObjects.map((option, i) => {
```
danparnella, line 153: Might as well make the same changes made to [Checkbox Group](https://github.com/LaunchPadLab/lp-components/pull/595/files/036cf6d2d2be75b107bfa164dea4bb8a8c5fe274#r1334902078) here as well:
- Removing the lint disabling
- Pulling up `key` out of the props spread
- Setting `key` to `option.value` instead of index

src/forms/labels/labeled-field.js
danparnella: Agreed. Since we are providing the ability now to set this to the correct element that would provide the correct aria role, I wouldn't think we'd need to specify one and should modify the test instead.

test/forms/inputs/checkbox-group.test.js
```diff
@@ -125,3 +125,21 @@ test('CheckboxGroup passes down props to children', () => {
   expect(checkbox).not.toHaveClass('custom-group-class')
   expect(checkbox).toHaveClass('custom-input-class')
 })
+
+test('DropdownCheckboxGroup sets menu no longer active when clicked outside', async () => {
```
danparnella, line 129: Since this isn't a real component anymore, should this be "CheckboxGroup with dropdown = true" (or similar)?

test/forms/inputs/checkbox-group.test.js
```diff
@@ -125,3 +125,21 @@ test('CheckboxGroup passes down props to children', () => {
...
+  const user = userEvent.setup()
+
+  render(<WrappedCheckboxGroup value={['1']} useDropdown />)
+
+  const select = screen.getByRole('button')
+  await user.click(select)
+
+  expect(select.nextSibling).toHaveClass('options', 'is-active')
+
+  const fieldset = screen.getAllByRole('group').at(0)
```
danparnella, line 139: Once we remove the `role` attribute from `LabeledField`, does this still work? I know creating it as a `fieldset` automatically gives it a role of group, but is that actually available to select by here?

src/forms/labels/labeled-field.js
danparnella: Just a heads up, once we remove that role, several other tests will need to be updated.

danparnella (changes requested):
Ok, I've gone through everything! Overall, I think this is a really good approach to solve the issue and provides a lot of flexibility. I agree with the comments Conor left and I had a few additional ones added here as well.

src/forms/inputs/radio-group.js
chawes13: Good idea. I'm happy to pull that out into a separate issue/PR if it helps with timing (wouldn't want that to happen without new tests, documentation, etc.)

.storybook/styles/components/_forms.scss
```diff
@@ -571,13 +553,8 @@ Custom Dropdown Select
     }
   }
 
-  fieldset.checkbox,
-  fieldset.CheckboxGroup {
-    width: 100% !important;
-
-    input {
```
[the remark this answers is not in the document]
jhp0621: Yes, this HTML structure is no longer true.

Before: https://share.zight.com/p9uY6mZ7
After: https://share.zight.com/geu1mX1W

We were unnecessarily wrapping the dropdown options in a `fieldset` and each option was also wrapped in a `fieldset`.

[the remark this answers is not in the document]
jhp0621: There already is a parent `fieldset` that renders the label legend

package.json
```diff
@@ -26,8 +26,7 @@
     "docs": "documentation build src/index.js -f md -o docs.md",
     "format": "prettier --write \"(src|test|stories|.storybook)/**/*.+(js|jsx|json|scss)\"",
     "lint": "eslint src --max-warnings=0",
-    "prepare": "husky install",
-    "prepublishOnly": "yarn run clean && yarn run build",
+    "prepare": "husky install && yarn run clean && yarn run build",
```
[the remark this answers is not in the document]
jhp0621: Yup 100%.

While working on figuring out how to add a private github url as a package dependency, I discovered that having babel compilation tasks under `prepublish` script does not work well if someone were to consume the repository directly via git protocol (cc: https://yarnpkg.com/advanced/lifecycle-scripts#prepublish). Although I image this use case would be rare, I still think it's good practice to place yarn build logic as part of prepare.

migration-guides/v.10.0.0.md
```diff
@@ -0,0 +1,53 @@
+# v10.0.0 Migration Guide
+
+This version contains the following breaking changes:
+
+1. The `DropdownCheckboxGroup` component is now `CheckboxGroup` with prop `dropdown`=`true`
+2. The `CheckboxGroup` and `RadioGroup` legends now rely on `visually-hidden` class styles to hide the label from the view
+
```
chawes13, line 7: I think we're missing a big breaking change here – `fieldset` no longer wraps each input! This will affect base styles in `_forms.scss`, as well as any **custom** rules that rely on this structure

migration-guides/v.10.0.0.md
```diff
@@ -0,0 +1,53 @@
...
+<DropdownCheckboxGroup input={inputProps} meta={{}} options={options} />
+
+// After:
+import { CheckboxGroup } from 'lp-components'
+
+<CheckboxGroup input={inputProps} meta={{}} options={options} dropdown={true} />
+```
+
+## 2. The `CheckboxGroup` and `RadioGroup` legends now rely on `visually-hidden` class styles to hide the label from the view. 
+To follow the accessibility guideline of `fieldset` having one, unique `legend` as its direct child, changes have been made in `CheckboxGroupLegend` and `RadioGroupLegend` to always render the legend in HTML and if `label=false` is passed into the `CheckboxGroup` and `RadioGroup` components, the class name "visually-hidden" is added to the legend in order to visually hide it. Please make sure the following class styles are included in your project's root stylesheet:
```
chawes13, line 39: `CheckboxGroupLegend` and `RadioGroupLegend` are not part of the public API (i.e., they aren't exported from this package). Referencing them directly may lead to confusion for developers. To mitigate this, I think you could just say `CheckboxGroup` and `RadioGroup`.

src/forms/helpers/dropdown-select.js
```diff
@@ -15,7 +15,7 @@ const defaultProps = {
   selectedValues: [],
 }
 
-// Wraps the `DropdownCheckboxGroup` component
+// Wraps the `CheckboxGroup` component: this is only when dropdown is true.
```
chawes13, line 18: 🍔 I think this would read more naturally as "Wraps the `CheckboxGroup` component when `dropdown` is set to true."

src/forms/inputs/checkbox-group.js
```diff
@@ -34,6 +36,7 @@ import {
  * @param {Object} meta - A `redux-form` [meta](http://redux-form.com/6.5.0/docs/api/Field.md/#meta-props) object
  * @param {Array} options - An array of checkbox values (strings, numbers, or key-value pairs)
  * @param {Object} [checkboxInputProps={}] - An object of key-value pairs representing props to pass down to all checkbox inputs
+ * @param {Boolean} [dropdown=false] - A boolean that determines whether the checkbox options are displayed in a dropdown container or not
```
chawes13, line 39: 🍔 "{Boolean} - A boolean..." is redundant. That being said, we're pretty inconsistent in this repo about that!

Here's how it looks in the documentation:
Image 2023-11-02 at 9 16 31 AM

src/forms/labels/labeled-field.js
```diff
@@ -72,10 +73,12 @@ const propTypes = {
   ...InputError.propTypes,
   children: PropTypes.node,
   hideErrorLabel: PropTypes.bool,
+  as: PropTypes.string,
```
chawes13, line 76: Do we want to be this permissive or are `div` and `fieldset` the only viable options?

src/forms/labels/labeled-field.js
```diff
@@ -88,23 +91,25 @@ function LabeledField({
...
   const { touched, invalid } = meta
   return (
-    <fieldset
-      className={classnames(className, {
+    <Wrapper
+      className={classnames(className, 'field-wrapper', {
         error: hasInputError({ touched, invalid }),
         disabled: rest.disabled,
       })}
+      data-testid="field-wrapper"
```
chawes13, line 105: Should this be dynamic based on the input's name? If you have a test with multiple inputs, it might be challenging to locate a particular container. Not sure there's a use case for that though 🤷

test/forms/inputs/checkbox-group.test.js
```diff
@@ -14,7 +17,7 @@ const WrappedCheckboxGroup = (props) => {
 
   const defaultProps = {
     input: {
-      name: 'test',
+      name: name,
```
chawes13, line 20: 🍔 shorthand assignment would work here

test/forms/inputs/checkbox-group.test.js
```diff
@@ -118,10 +121,86 @@ test('CheckboxGroup passes down props to children', () => {
...
+
+  const fieldset = screen.getAllByRole('group').at(0)
+
+  user.click(fieldset)
+  await waitFor(() => {
+    expect(select.nextSibling).not.toHaveClass('is-active')
+  })
+})
+
+test('when no custom required indicator provided, do not show required indicator', () => {
```
chawes13, line 188: There is no outer `describe` block in this test file, so this test case (and the remaining) should be prefixed with `CheckboxGroup`

test/forms/inputs/icon-input.test.js
```diff
@@ -9,9 +9,9 @@ const props = { input: { name, value, onChange }, meta: {}, icon: 'foo' }
 
 test('IconInput adds class "icon-label" to surrounding fieldset', () => {
```
chawes13, line 10: I believe the test description needs to be rewritten not to directly reference `fieldset`

test/forms/labels/labeled-field.test.js
```diff
@@ -10,9 +10,9 @@ test('wraps children in fieldset', () => {
       <Wrapped />
```
chawes13, line 10: I believe the test description needs to be rewritten to not directly reference `fieldset`

chawes13: Could we add a test that explicitly tests the `as="fieldset"` case? E.g.,

`expect(screen.getByRole('group')).toBeInTheDocument()`

test/forms/inputs/radio-group.test.js
```diff
@@ -131,3 +142,58 @@ test('RadioGroup passes down props to children', () => {
     expect(el).toHaveAttribute('data-test', 'true')
   })
 })
+
+test('when no custom required indicator provided, do not show required indicator', () => {
```
chawes13, line 146: Can you prefix this test (and the following) with `RadioGroup` to match the rest of the descriptions in this file?

test/forms/inputs/radio-group.test.js
```diff
@@ -80,11 +86,12 @@ test("RadioGroup does not have a legend when label is `false`", () => {
     label: false,
   }
   render(<RadioGroup {...props} />)
-  expect(screen.queryByRole('group', { name: 'Same Name' })).not.toBeInTheDocument()
+  expect(
+    screen.queryByRole('group', { name: formattedName })
+  ).toBeInTheDocument()
```
chawes13, line 91: Can we add an assertion about `visually-hidden` being applied?

test/forms/labels/labeled-field.test.js
jhp0621: @chawes13 I'm actually not sure what this first test is accomplishing... I'm merely looking for an element with the test-id I assigned and testing that it exists (without 'error' class). Should we keep it?

chawes13: Yes we should keep it. We have several components **and** styling that rely on LabeledField returning a containing element (div or fieldset), as opposed to a fragment

chawes13: Now that you're adding `'field-wrapper'` to this container, that'd be a useful thing to assert on (especially since our styles depend on it!)

src/forms/labels/labeled-field.js
jhp0621: decided to ditch `data-testid` and instead use `{ container }` from `render` results as you suggested! https://testing-library.com/docs/react-testing-library/api/#container-1

src/forms/labels/labeled-field.js
jhp0621: I think we should be strict with what's allowed and if we ever receive requests re: expanding the options (for whatever reason that I can't think of atm), we can address then!

jhp0621: cc: @danparnella

migration-guides/v.10.0.0.md
jhp0621: I also added an additional call out about `LabeledField` itself (bullet point 2) since that is an exported component that can be used on its own.. thoughts?

src/forms/inputs/checkbox-group.js
```diff
@@ -118,22 +150,28 @@ function CheckboxGroup(props) {
     return function (checked) {
       // Add or remove option value from array of values, depending on whether it's checked
       const newValueArray = checked
-        ? addToArray([option.value], value)
+        ? addToArray(value, [option.value])
```
jhp0621, line 153: Flipped the order so that the new selection shows at the end of the input, rather than at the beginning

src/forms/helpers/dropdown-select.js
```diff
@@ -8,27 +8,38 @@ const propTypes = {
...
-function DropdownSelect({ children, className, selectedValues }) {
+function DropdownSelect({ children, className, selectedValues, options }) {
   const [expanded, toggleExpanded] = useToggle()
 
   return (
     <OutsideClickHandler onOutsideClick={() => toggleExpanded(false)}>
-      <div className="dropdown-select">
+      <div
+        className={classnames('dropdown-select', {
+          'is-active': expanded,
```
jhp0621, line 34: It seems to be a good idea to put "is-active" class name at the top level for styling purpose (e.g., coloring the border of the input when the dropdown menu is open)

jhp0621: cc: https://github.com/LaunchPadLab/client-template/pull/456/files#diff-102f9f47138cbbdf4cafcec5b8d376a8a3e508d085c7a18668ca5704039fc207R271

test/forms/inputs/checkbox-group.test.js
```diff
@@ -118,10 +121,87 @@ test('CheckboxGroup passes down props to children', () => {
...
-  const checkboxGroup = screen.getByRole('group', { name: 'Test Group' })
+  const checkboxGroup = screen.getByRole('group', { name: formattedName })
   const checkbox = screen.getByRole('checkbox')
 
   expect(checkboxGroup).toHaveClass('custom-group-class')
   expect(checkbox).not.toHaveClass('custom-group-class')
   expect(checkbox).toHaveClass('custom-input-class')
 })
+
+test('CheckboxGroup with dropdown = true adds value to array when unselected option clicked', async () => {
```
danparnella, line 132: Looks like Travis is failing on this test

danparnella (changes requested):
I'll try to do one more holistic review, but I had a few comments on the changes since my last review.

src/forms/inputs/checkbox-group.js
```diff
@@ -145,8 +183,8 @@ function CheckboxGroup(props) {
               ...checkboxInputProps,
```
danparnella, line 183: Not a new issue, but can you resolve the accessibility issue for the "with input props specified" story? Looks like we'll need to set the `id` here like we're doing in `RadioGroup` ([line](https://github.com/LaunchPadLab/lp-components/pull/595/files#diff-5c2fdf596caab0ddfe40eef9596137b37bd03591ab1496e366d363d9ad2e9ac2R164)).

src/forms/labels/labeled-field.js
```diff
@@ -72,10 +73,12 @@ const propTypes = {
...
   hideErrorLabel: PropTypes.bool,
+  as: PropTypes.oneOf(['div', 'fieldset']),
 }
 
 const defaultProps = {
   hideErrorLabel: false,
+  as: 'div',
 }
 
 function LabeledField({
```
danparnella, line 84: Again, not new, but there are a few stories for this component with accessibility issues that can be seen in the storybook (form elements not having labels).

danparnella (changes requested):
I found a few storybook accessibility violations for the components we've touched that we should be able to resolve pretty quickly.

src/forms/inputs/checkbox-group.js
chawes13: IMO let's save this for a fast-follow patch release. I'd prefer not to delay this release any longer

src/forms/inputs/checkbox-group.js
chawes13: Good catch! We should add a note about this in the migration guide

src/forms/labels/labeled-field.js
danparnella: 👍 Issue created: https://github.com/LaunchPadLab/lp-components/issues/632

chawes13:
Removed my review to take it out of my queue, please retag me when y'all are comfortable with the PR and ready for me to take a look!

danparnella (approved):
Updates resolved everything I saw previously. LGTM! Just need to resolve the version number conflict.

@chawes13 Ready for you to double-check!

src/forms/helpers/dropdown-select.js
```diff
@@ -46,8 +57,9 @@ DropdownSelect.propTypes = propTypes
 
 DropdownSelect.defaultProps = defaultProps
 
-function getLabel(values) {
-  return values.length ? values.join(', ') : 'None'
+function getLabel(values, options) {
+  const labels = values.map((v) => options.find((o) => o.value === v).key)
+  return values.length ? labels.join(', ') : 'None'
```
chawes13, line 62: While `labels.length` should always equal `values.length`, it's safer to have the guard on `labels.length`

chawes13 (approved):
Lovely 👏 

Confirmed test coverage remains at 100%! 🚀 
Image 2024-02-12 at 12 01 05 PM

PR #600, opened 2023-08-04
RTL migration: indicators

chawes13:
Updates the remaining Indicator component test to use React Testing Library

## Author Checklist

- [x] Add unit test(s)
- [x] Assign dev reviewer

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,7 @@
     "@storybook/builder-webpack5": "^6.5.14",
     "@storybook/manager-webpack5": "^6.5.14",
     "@storybook/react": "^6.4.22",
+    "@testing-library/dom": "^9.3.1",
     "@testing-library/jest-dom": "^5.16.5",
     "@testing-library/react": "^12.1.5",
     "@testing-library/user-event": "^14.4.3",
--- a/src/indicators/flash-message-container.js
+++ b/src/indicators/flash-message-container.js
@@ -47,7 +47,6 @@ function FlashMessageContainer({ messages, limit, onDismiss, ...rest }) {
         return (
           <FlashMessage
             key={message.id}
-            message={message}
             isError={message.isError}
             {...rest}
             {...message.props}
--- a/test/indicators/flash-message-container.test.js
+++ b/test/indicators/flash-message-container.test.js
@@ -1,70 +1,74 @@
 import React from 'react'
-import { mount } from 'enzyme'
+import { render, screen } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import { within } from '@testing-library/dom'
 import { FlashMessageContainer } from '../../src'
 
 const successMessage = {
   id: '0',
   message: 'Success!',
   isError: false,
-  props: {},
+  props: {
+    'data-testid': 'success'
+  },
 }
 const failureMessage = {
   id: '1',
   message: 'Failure!',
   isError: true,
-  props: {},
+  props: {
+    'data-testid': 'failure'
+  },
 }
 
-test('FlashMessageContainer displays all provided redux-flash messages', () => {
-  const wrapper = mount(
+test('FlashMessageContainer displays all provided flash messages', () => {
+  render(
     <FlashMessageContainer messages={[successMessage, failureMessage]} />
   )
-  expect(wrapper.find('div.flash-message.success').exists()).toBe(true)
-  expect(
-    wrapper.find('div.flash-message.success > p').first().contains('Success!')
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -71,6 +71,7 @@
     "@storybook/builder-webpack5": "^6.5.14",
     "@storybook/manager-webpack5": "^6.5.14",
     "@storybook/react": "^6.4.22",
+    "@testing-library/dom": "^9.3.1",
```
chawes13, line 74: We're using this directly to import `within`

https://testing-library.com/docs/dom-testing-library/api-within/

src/indicators/flash-message-container.js
```diff
@@ -47,7 +47,6 @@ function FlashMessageContainer({ messages, limit, onDismiss, ...rest }) {
         return (
           <FlashMessage
             key={message.id}
-            message={message}
```
chawes13, line 50: This was an accidental regression that I introduced in my work last year. I don't think it should be considered a breaking change because it's a bug and not documented (or supported by prop types)

test/indicators/flash-message-container.test.js
```diff
@@ -1,70 +1,74 @@
...
+  },
 }
 
-test('FlashMessageContainer displays all provided redux-flash messages', () => {
-  const wrapper = mount(
+test('FlashMessageContainer displays all provided flash messages', () => {
+  render(
     <FlashMessageContainer messages={[successMessage, failureMessage]} />
   )
-  expect(wrapper.find('div.flash-message.success').exists()).toBe(true)
```
chawes13, line 22: I didn't migrate this assertion because I think it's actually testing the `FlashMessage` component, which we have separate tests for! If the class name was changed, I don't think this test should fail (but the spec in `flash-message.test.js` should)

test/indicators/flash-message-container.test.js
```diff
@@ -1,70 +1,74 @@
...
-  expect(
-    wrapper.find('div.flash-message.success > p').first().contains('Success!')
-  ).toBe(true)
-  expect(wrapper.find('div.flash-message.failure').exists()).toBe(true)
```
chawes13, line 26: Same comment as above

test/indicators/flash-message.test.js
```diff
@@ -1,11 +1,11 @@
...
+import { render, screen } from '@testing-library/react'
 import { FlashMessage } from '../../src/'
 
 test('FlashMessage only shows dismiss button when callback is provided', () => {
-  const wrapper = mount(<FlashMessage>Success!</FlashMessage>)
-  expect(wrapper.find('button.dismiss').exists()).toBe(false)
-  const dismissWrapper = mount(
+  render(<FlashMessage>Success!</FlashMessage>)
+  expect(screen.queryByRole('button')).not.toBeInTheDocument()
+  render(
```
chawes13, line 8: Should I be using `rerender` here?

test/indicators/flash-message-container.test.js
josiasds: That makes sense 👍

test/indicators/flash-message.test.js
josiasds: Not in this case. `rerender` should be used for testing the result of updating a prop, and it looks like we're just testing two different cases (when `onDismiss` is provided vs. not provided).

test/indicators/flash-message.test.js
```diff
@@ -27,14 +27,17 @@ test('FlashMessage dismiss button includes label for screenreaders', () => {
       Success!
     </FlashMessage>
   )
-  expect(dismissWrapper.find('button.dismiss').prop('aria-label')).toBeDefined()
+  expect(screen.getByRole('button')).toHaveAttribute('aria-label')
```
josiasds, line 30: You can test the `aria-label` text directly here:
```suggestion
  expect(screen.getByLabelText('Dismiss'))
```

josiasds (approved):
This looks great!

test/indicators/flash-message.test.js
chawes13: Thanks! I tweaked this slightly to assert `toBeInTheDocument()`. I don't love the empty chain on `expect` because if the API were to change in the testing library, this test would still pass (`expect(null)`, `expect(undefined)`, and `expect(false)` all pass!).

josiasds: That's right! I only included the first part of the assertion in the code suggestion, my bad.

PR #601, opened 2023-08-04
RTL migration: modal

chawes13:
Updates modal test to use RTL

## Author Checklist

- [x] Add unit test(s)
- [x] Assign dev reviewer

### DIFF
--- a/test/modal.test.js
+++ b/test/modal.test.js
@@ -1,7 +1,9 @@
 import React from 'react'
-import { mount } from 'enzyme'
+import { render, screen } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
 import { Modal } from '../src/'
 import { noop } from 'lodash'
+
 
 describe('Modal', () => {
   beforeEach(() => {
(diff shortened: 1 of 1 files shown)


test/modal.test.js
```diff
@@ -20,38 +22,42 @@ describe('Modal', () => {
...
   test('adds additional overlay class string to default overlay class', () => {
-    const wrapper = mount(
+    render(
       <Modal onClose={noop} overlayClassName="custom-overlay" />
     )
-    expect(wrapper.find('.modal-fade-screen.custom-overlay').exists()).toEqual(
-      true
-    )
+
+    const overlay = screen.getByRole('dialog').parentElement
```
chawes13, line 55: I don't love having to use `parentElement`, but that seemed to be the easiest option. Let me know if you have any ideas

test/modal.test.js
```diff
@@ -79,45 +83,46 @@ describe('Modal', () => {
...
           preventClose={true}
           onClose={onClose}
-          overlayClassName={overlayClass}
         />
       )
-      wrapper.find('.' + overlayClass).simulate('click')
+      const overlay = screen.getByRole('dialog').parentElement
+      await user.click(overlay)
       expect(onClose).not.toHaveBeenCalled()
-      expect(wrapper.find('.modal-content').exists()).toEqual(true)
```
chawes13, line 112: This assertion is actually a bit of a false flag – the test wasn't changing `isOpen` state, so even if `onClose` was called, this would still pass. We don't need it.

test/modal.test.js
josiasds: I think it makes sense in this case. We're talking about an overlay element, so it's naturally a wrapper to the modal. Another option would be to add a testId to the overlay element, but I don't think it would add any value here.

test/modal.test.js
```diff
@@ -79,45 +83,46 @@ describe('Modal', () => {
...
+    test('does not close by escape key', async () => {
       const onClose = jest.fn()
-      const wrapper = mount(<Modal preventClose={true} onClose={onClose} />)
-      wrapper.find('.modal-content').simulate('keydown', { keyCode: 27 })
+      const user = userEvent.setup()
+      render(<Modal preventClose={true} onClose={onClose} />)
+
+      await user.keyboard('{Escape}')
       expect(onClose).not.toHaveBeenCalled()
     })
```
josiasds, line 103: Should we add tests for the default case, that it does close on Esc and on overlay click? I realize that it's not part of the original tests but it would make them more reliable.

josiasds (approved):
Looks good!

test/modal.test.js
chawes13: Yeah, that makes sense to me. Will add now

PR #607, opened 2023-08-22
RTL migration: labels

chawes13:
Migrates the label tests from Enzyme to RTL.

Creating a new PR to limit the broadcast of notifications.


test/forms/labels/error-label.test.js
```diff
@@ -1,15 +1,22 @@
...
+test('passes extra props to span element', async () => {
+  let count = 0
+  const onClick = () => count++
+  const user = userEvent.setup()
+
+  render(<ErrorLabel onClick={onClick}>Error!</ErrorLabel>)
+
+  await user.click(screen.getByText('Error!'))
+
+  expect(count).toBe(1)
```
josiasds, line 21: This is interesting. I haven't seen that before, but it seems to cover the test case.
Wouldn't passing a data attribute be simpler though?

test/forms/labels/input-error.test.js
```diff
@@ -1,62 +1,66 @@
...
-      onFancyClick={onClick}
       error="Foo"
+      test="test"
       touched
       invalid
     />
   )
-  expect(wrapper.props().onFancyClick).toBe(undefined)
+
+  expect(screen.getByText('Foo')).not.toHaveAttribute('touched', 'invalid')
```
josiasds, line 58: I'm not sure I understand this test. `test` is the invalid prop being passed, but we're checking `touched` and `invalid`. Am I missing something?

test/forms/labels/input-label.test.js
```diff
@@ -1,93 +1,109 @@
...
+import userEvent from '@testing-library/user-event'
 
 const name = 'person.firstName'
 const formattedName = 'First Name'
 
 test('when label is false - does not render a label', () => {
-  const wrapper = mount(<InputLabel name={name} label={false} />)
-  expect(wrapper.find('label').exists()).toEqual(false)
+  render(<InputLabel name={name} label={false} />)
+  expect(screen.queryByText(formattedName)).not.toBeInTheDocument()
```
josiasds, line 11: I'm not sure this test works as intended. It still would pass if a label was rendered. It just tests the variable, which is never passed anyway.

test/forms/labels/error-label.test.js
chawes13: Yeah. This was written prior to me taking over the PR. I can modify it

test/forms/labels/input-error.test.js
chawes13: Those are also not valid attributes, right?

chawes13: Oh, I see yeah it doesn't test much 😆

test/forms/labels/input-label.test.js
chawes13: What do you mean it just tests the variable? When I remove `label={false}` the test fails

josiasds: Oh I just noticed that `InputLabel` converts the text using `convertNameToLabel` internally. So that's why you're testing against this variable.

josiasds: Testing against `formattedName` didn't make sense to me because you never passed this value anyway. I see the connection now.

chawes13: Yes that is true, if you passed in a different label the test would need to be updated to search for that label

josiasds: Anyway, since you're strictly testing for `label={false}` in this test, I think we're fine.

josiasds (approved):

PR #621, opened 2023-11-09
Fix flaky aria-disabled tests

josiasds:
@chawes13 found a flaky test when reviewing the [RTL Cheat Sheet](https://www.notion.so/launchpadlab/React-Testing-Library-RTL-Cheatsheet-82af6c02b19f4a9eb5cde8f64901a4d0?pvs=4).

We were only checking for the existence of the `aria-disabled` attribute but not its value. If `aria-disabled` were `false`, the test would still pass, and that's not the desired outcome.

## Author Checklist

- [x] Add unit test(s)
- [x] Assign dev reviewer

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@launchpadlab/lp-components",
-  "version": "9.2.1",
+  "version": "9.2.2",
   "engines": {
     "node": "^18.12"
   },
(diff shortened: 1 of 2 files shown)


chawes13 (approved):
I noticed this too! You beat me to it. Thanks a bunch 🙏

PR #637, opened 2024-03-12
Allow Class Specification on TableColumn

### LINKED ISSUE #633: Allow Class Specification on TableColumn
## Feature Description

By default in the client template, tables use `table-layout: fixed`. By default, this configuration makes all columns an equal size (see: https://css-tricks.com/almanac/properties/t/table-layout/).

If a design requires some table columns to be larger than others, specification of a size on the "first table row" (typically, one or more `<th>` elements) is required. While it is possible to provide a custom `headerComponent` to each sized column to resolve this issue, creation of a component to be able to set a class seems like overkill. Rather, it would be easier to simply pass a class on the `<TableColumn>` component and have the class be propagated to the `<th>` element in the `<TableHeader>`.

## Suggested Solution

I believe this can be resolved simply by:
1. Pulling  an explicit `className` prop out of the `column` prop in the `TableHeader` definition.
2. Passing the resulting class  to the rendered `<th>` element's `className`.

That is:
````
function TableHeader({
  column: { className, name, label, disabled },
  sortPath,
  ascending,
  onClick,
}) {
  const active = sortPath === name
  const arrowClass = getArrowClass(active, ascending)
  return (
    <th
      onClick={onClick}
      className={classnames(arrowClass, { sortable: !disabled, className })}
    >
      {label || startCase(name)}
    </th>
  )
}
````

## Alternatives Considered / Existing Workarounds
Creation of a specific header component instance for each column on which a size specification is required.

### PR DESCRIPTION
**Pull request by Ramy-Huffman-LPL**
This is in reference to an [enhancement](https://github.com/LaunchPadLab/lp-components/issues/633) to add to a `Table` and pass a custom classname. 

## Author Checklist

- [x] Add unit test(s)
- [x] Assign dev reviewer

### DIFF
--- a/test/tables/sortable-table.test.js
+++ b/test/tables/sortable-table.test.js
@@ -476,4 +476,15 @@ describe('SortableTable', () => {
     )
     expect(screen.getByText('My Table')).toHaveClass('custom-caption')
   })
+
+  test('Column with custom className is propagated to TableHeader', () => {
+    render(
+      <SortableTable data={tableData}>
+        <Column name="name" className="foo" />
+      </SortableTable>
+    )
+
+    const header = screen.getByText('Name').closest('th')
+    expect(header).toHaveClass('foo')
+  })
 })
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@launchpadlab/lp-components",
-  "version": "10.0.0",
+  "version": "10.1.0",
   "engines": {
     "node": "^18.12 || ^20.0"
   },
--- a/src/tables/components/table-header.js
+++ b/src/tables/components/table-header.js
@@ -12,7 +12,7 @@ const propTypes = {
 }
 
 function TableHeader({
-  column: { name, label, disabled },
+  column: { className, name, label, disabled },
   sortPath,
   ascending,
   onClick,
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Comment by chawes13** [comment]
@Ramy-Huffman-LPL I'm going to pull myself off of the review until this PR is converted from a draft. LMK if you have any questions in the meantime!

**Comment by chawes13** [comment]
@Ramy-Huffman-LPL FYI if you use keywords like "Resolves " in front of your link to the issue, it will not only link it to the Development section of this PR, it will also autoclose it when the PR is merged!

Here's more info: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

File: test/tables/sortable-table.test.js [hunk 1]
```diff
@@ -476,4 +476,15 @@ describe('SortableTable', () => {
...
   })
+
+  test('Column with custom className is propagated to TableHeader', () => {
+    render(
+      <SortableTable data={tableData}>
+        <Column name="name" className="foo" />
+      </SortableTable>
+    )
+
+    const header = screen.getByText('Name').closest('th')
```
chawes13, line 487: Why do you need `closest` here? When I printed out the value for `screen.getByText('Name')`, it was the header itself 🤔

chawes13 (changes requested):

chawes13:
@Ramy-Huffman-LPL will you be able to pick this back up this quarter?

Issue #631, opened 2024-01-05
Pull out legend into separate component


danparnella:
## Feature Description
- In the new updates coming in #595, both the Checkbox Group and Radio Group components use essentially the same legend component.

## Suggested Solution
- Pull that legend component out into its own component so that it can be reused in the library as well as become available to be used separately on projects if needed (from https://github.com/LaunchPadLab/lp-components/pull/595#discussion_r1339083065).

PR #640, opened 2024-03-26
Create reusable FieldSetLegend component

jhp0621:
https://github.com/LaunchPadLab/lp-components/issues/631

## Author Checklist

- [x] Add unit test(s)
- [x] Assign dev reviewer


src/forms/labels/field-set-legend.js
```diff
@@ -0,0 +1,88 @@
...
+ * - Else If the `label` prop is set to a string, the label will display that text
+ * - Otherwise, the label will be set using the `name` prop.
+ *
+ * 
+ * @name FieldSetLegend
+ * @type Function
+ * @param {String} name - The name of the associated group
+ * @param {String} [hint] - A usage hint for the associated input
+ * @param {String|Boolean} [label] - Custom text for the legend
+ * @param {Boolean} [required] - A boolean value to indicate whether the field is required
```
chawes13, line 24: This should have a default. You can communicate that with `=` within the brackets

src/forms/labels/field-set-legend.js
```diff
@@ -0,0 +1,88 @@
...
+ * @param {String} [requiredIndicator] - Custom character to denote a field is required (optional, default `''`)
```
chawes13, line 25: You don't need to specify `optional, default` in the description. `optional` is communicated already using the brackets (i.e., `[]`) and you can set the default by using `=''`.

Here's an example: [input](https://github.com/LaunchPadLab/lp-components/blob/052921a7cb7642867ce76ef4ea12a5f2a62b9144/src/forms/inputs/range-input.js#L25) and  [output](https://github.com/LaunchPadLab/lp-components/blob/main/docs.md#rangeinput)

src/forms/labels/field-set-legend.js
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ *
+ * A legend representing a caption for the content of its parent field set element
+ *
+ * This component must be used as a direct child and the only legend of the <fieldset> element that groups related controls
+ * 
+ *
+ * The text of the legend is set using the following rules:
+ * - If the `label` prop is set to `false`, the legend is hidden visually
```
chawes13, line 14: Let's specify that it's hidden visually via a class and that its the consumer's responsibility to make sure their styling rules respect the `visually-hidden` class

stories/forms/labels/fieldset-legend.story.js
```diff
@@ -0,0 +1,24 @@
...
+
+storiesOf('FieldSetLegend', module)
+  .add('with default label', () => <FieldSetLegend name="nameOfInput" />)
+  .add('with custom label', () => (
+    <FieldSetLegend name="nameOfInput" label="Custom Label" />
+  ))
+  .add('with no label', () => (
+    <FieldSetLegend name="nameOfInput" label={false} />
+  ))
+  .add('with required true default indicator', () => (
```
chawes13, line 13: This makes it read like there is a required default indicator and I could see there being some confusion about nothing showing up since by default we do not show an indicator

src/forms/labels/field-set-legend.js
```diff
@@ -0,0 +1,88 @@
...
+  hint: PropTypes.string,
+}
+
+const defaultProps = {
+  children: null,
+  hint: '',
+  label: '',
+  requiredIndicator: '',
+  className: '',
+}
```
chawes13, line 60: Missing default for `required`

chawes13 (commented):
Lookin' good! A few documentation nits. Tests passed at 100% coverage 🚀

src/forms/labels/field-set-legend.js
```diff
@@ -0,0 +1,88 @@
...
+
+function FieldSetLegend({
```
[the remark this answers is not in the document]
chawes13: Source? 😅😅😅😅

MDN disagrees (but only with one usage) https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset

chawes13:
@jhp0621 where's this one stand?

src/forms/labels/field-set-legend.js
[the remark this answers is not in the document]
chawes13: ChatGPT agrees with Danny. Should probably stick with what the ai overlords tell us (since they'll be maintaining this in a bit)
Image 2024-04-30 at 1 10 09 PM

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@launchpadlab/lp-components",
-  "version": "10.0.1",
+  "version": "10.1.1",
```
chawes13, line 3: When you increment a new minor or major version, all of the other numbers are reset to 0. 

https://semver.org/#spec-item-7 & https://semver.org/#spec-item-8

chawes13 (changes requested):
Code looks good, but let's switch the file and function name to "fieldset".

